PR #66, opened 2026-04-07
feat: add recovery hints and related tools to error responses

junho226:
## Summary

- Add `recovery_hints` and `related_tools` fields to error responses so LLMs can self-recover from errors
- Auto-detect error domain (command, server, file, user, alert) from message, tool name, and endpoint
- Enrich error responses transparently via `with_error_handling` decorator — no tool function changes needed

## Pipeline Results
- Tests: PASS (494 tests, 0 failures)
- Review: APPROVED (0 critical, 2 informational)
- Auto-fix: SKIPPED (no items)

## Changes
- `utils/recovery_hints.py` (new): hint registry mapping `(status_code, domain)` → hints + related tools, `enrich_error_response()` function
- `utils/decorators.py`: `with_error_handling` now post-processes all error responses with recovery hints
- `tests/test_recovery_hints.py` (new): 35 tests covering domain detection, hint lookup, enrichment logic

## Error response example (before → after)
```python
# Before
{"status": "error", "message": "...", "status_code": 403}

# After
{"status": "error", "message": "...", "status_code": 403,
 "recovery_hints": ["Check if the API token has 'command_execute' ACL permission.", ...],
 "related_tools": ["list_command_acls", "create_command_acl"]}
```

## Testing
- [x] All 494 tests pass
- [x] 35 new tests for recovery hints module
- [x] No in-branch test failures

Closes #61


eunyoung14 (dismissed):

eunyoung14 (dismissed):

utils/recovery_hints.py
```diff
@@ -0,0 +1,244 @@
...
+    if '/api/servers/' in ep_lower:
+        return 'server'
+    if '/api/iam/' in ep_lower:
+        return 'user'
+
+    return 'general'
+
+
+# Hint registry: (status_code, domain) -> {recovery_hints, related_tools}
+_HINT_REGISTRY: dict[tuple[int, str], dict[str, list[str]]] = {
```
jisung-02, line 6: _HINT_REGISTRY is sandwiched between _detect_error_domain and _parse_status_code. Module-level constants typically go above the function definitions (PEP 8 convention: imports → constants → functions).

utils/recovery_hints.py
```diff
@@ -0,0 +1,244 @@
...
+    # 500 - Server error
+    (500, 'general'): {
+        'recovery_hints': [
+            'The server encountered an internal error. Try again in a moment.',
+            'If the issue persists, check server health.',
+        ],
+        'related_tools': [],
+    },
+    # Timeout
+    (0, 'general'): {
```
jisung-02: There doesn't seem to be any path that hits the `(0, 'general')` timeout entry (line 133) in the current code. (`status_code = response.get('status_code') or response.get('error_code')` falls back to `error_code` when `status_code` is 0.)

utils/recovery_hints.py
junho226: Fixed in a0e7e18. Moved `_HINT_REGISTRY` above the function definitions to follow the PEP 8 convention (imports → constants → functions).

utils/recovery_hints.py
junho226: Fixed in a0e7e18. You're right — `http_client.py` returns timeout responses as `{'error': 'Timeout', 'message': '...'}` with no `status_code`/`error_code`, so `_parse_status_code(None)` returns `None` and `get_recovery_hints` short-circuits at the early return before it can ever reach `(0, 'general')`. Removed the dead entry; we can revisit timeout-specific hints if/when timeout responses gain an explicit status code.

jisung-02 (approved):

PR #74, opened 2026-04-30
fix: restrict wheel to runtime-only files

### LINKED ISSUE #73: fix: unnecessary files included in wheel package
## Problem

The published wheel contains files that should not be distributed to end users.

**Root cause:** `pyproject.toml` sets `packages = ["."]`, which tells hatchling to include the entire root directory. Additionally, `MANIFEST.in` exists in the repo but has never been effective — it is a setuptools concept and is ignored by hatchling.

## Files incorrectly included in the wheel

- `tests/` — entire test suite (25+ files)
- `docs/` — developer documentation (8 files)
- `CLAUDE.md` — internal AI assistant instructions
- `.github/` — CI workflows and Copilot instructions
- `uv.lock` — 156 KB lock file
- `Dockerfile`, `.dockerignore`, `.env.example`
- `.pre-commit-config.yaml`, `.python-version`, `.rules`
- `pytest.ini`, `MANIFEST.in`, `CHANGELOG.md`, `CONTRIBUTING.md`, `LOGGING.md`

## Verification

Build the wheel locally and inspect:

```bash
uvx --from build pyproject-build --wheel --outdir /tmp/alpacon-build .
unzip -l /tmp/alpacon-build/alpacon_mcp-*.whl
```

## Fix

1. Update `[tool.hatch.build.targets.wheel]` in `pyproject.toml` to explicitly specify only the packages and files needed at runtime
2. Delete `MANIFEST.in` — it has never had any effect on this project

jisung-02:
## Summary

- Replace `packages=["."]` with explicit `include` list in hatchling wheel config — previously the entire repo root was bundled into the wheel
- Use recursive glob (`**/*.py`) for `tools/` and `utils/` to be future-proof against subdirectories
- Delete `MANIFEST.in` which was a setuptools artifact and has never had any effect on hatchling builds

## Verification

Built wheel locally and verified:
- 41 files total, all runtime-only
- No `tests/`, `docs/`, `CLAUDE.md`, `uv.lock`, `.github/`, etc.
- Entry points (`alpacon-mcp`, `alpacon-mcp-sse`, `alpacon-mcp-http`) confirmed working
- Live API call to dev/alpacax workspace successful (HTTP 200)

Closes #73


pyproject.toml
```diff
@@ -69,7 +69,17 @@ dev = [
 ]
 
 [tool.hatch.build.targets.wheel]
-packages = ["."]
+include = [
+    "/__init__.py",
+    "/py.typed",
+    "/server.py",
+    "/main.py",
```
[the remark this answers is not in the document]
jisung-02: Thanks for the review! The license text is already embedded in the wheel's `.dist-info/METADATA` file because `pyproject.toml` uses the inline form `license = { text = "MIT" }`. Hatchling writes this directly into the wheel metadata, so MIT redistribution requirements are satisfied without a separate `LICENSE` file entry in the `include` list.

If we ever switch to `license = { file = "LICENSE" }`, we'd need to add `/LICENSE` here — but for now this is fine.

pyproject.toml
```diff
@@ -69,7 +69,17 @@ dev = [
 ]
 
 [tool.hatch.build.targets.wheel]
-packages = ["."]
+include = [
+    "/__init__.py",
```
junho226, line 73: Suggestion: neither `/__init__.py` nor `/py.typed` exists at the repo root today. Hatchling silently ignores missing patterns, so this works, but it leaves aspirational entries in the config. If you intend to add a top-level package marker or PEP 561 marker later, fine; otherwise consider dropping these two lines so the include list mirrors what actually ships.

junho226 (approved):
## Summary

Replaces hatchling's overly-inclusive `packages = ["."]` (which was bundling the entire repo root into the wheel) with an explicit `include` allowlist. Uses recursive globs (`tools/**/*.py`, `utils/**/*.py`) to future-proof against new subpackages. Deletes dead `MANIFEST.in` (a setuptools artifact; hatchling ignores it).

## Verdict: APPROVED

## Suggestions

1. `/__init__.py` and `/py.typed` are listed in `include` but neither file exists at the repo root. Hatchling silently ignores missing patterns, so this is harmless today, but it suggests aspirational future state. If they're not landing in this PR, consider dropping them; otherwise they become misleading config drift.
2. Optional: a small CI step (`python -m build --wheel && unzip -l dist/*.whl`) asserting expected file count would prevent silent regressions if someone reintroduces a broad include pattern in the future.

## Highlights

- Correctly identifies that `MANIFEST.in` does nothing for hatchling — clean removal
- Explicit allowlist is safer than implicit `packages = ["."]`
- Author verified the wheel contents (41 runtime-only files, no `tests/`, `CLAUDE.md`, `uv.lock`, `.taskmaster/`, etc.)
- Live workspace API call verified the entry points still work post-build

## Checklist

- [x] Manual verification covered (wheel inspection + live API call)
- [x] No security vulnerabilities introduced
- [x] No documentation updates needed

pyproject.toml
jisung-02: Both files exist at the repo root — `/__init__.py` and `/py.typed` are present and verified locally. The entries are not aspirational; they reflect what actually ships.

junho226 (approved):
Confirmed: `/__init__.py` and `/py.typed` exist at the repo root — my earlier concern was based on incorrect inspection. The include list reflects what actually ships.

LGTM 👍

PR #85, opened 2026-04-30
Potential fix for code scanning alert no. 14: Workflow does not contain permissions

jisung-02:
Potential fix for [https://github.com/alpacax/alpacon-mcp/security/code-scanning/14](https://github.com/alpacax/alpacon-mcp/security/code-scanning/14)

Add an explicit `permissions` block to the workflow with least privilege required for current steps.  
Best fix here: define workflow-level permissions right under `on:` (or under `name:`), setting:

- `contents: read`

This preserves functionality (`actions/checkout` still works) while constraining token scope for all jobs unless overridden.  
Edit only `.github/workflows/test.yml`, inserting the block before `jobs:`.

No imports, methods, or dependencies are needed.

_Suggested fixes powered by Copilot Autofix. Review carefully before merging._


.github/workflows/test.yml
```diff
@@ -7,6 +7,9 @@ on:
   pull_request:
     branches: [ "main" ]
 
+permissions:
```
junho226, line 10: Suggestion: This minimum is correct for the current workflow. If future steps need to upload coverage, comment on PRs, or push artifacts, remember to scope additional permissions at the *job* level (e.g. `pull-requests: write`) rather than relaxing the workflow-level default — keeping the principle of least privilege at the top level.

junho226 (dismissed):
## Summary

This is a Copilot Autofix that adds an explicit workflow-level `permissions: contents: read` declaration to `.github/workflows/test.yml`, scoping down the default `GITHUB_TOKEN` to the minimum required by the current steps.

## Verdict

**APPROVED** — accurate, minimal-risk change that satisfies the code scanning alert.

## Notes

- Workflow only performs `actions/checkout@v4`, dependency install, ruff/mypy, and pytest. None of these require write scopes, so `contents: read` is correct and sufficient.
- No regression risk: no step relies on the wider default scope.
- Aligned with GitHub Actions hardening guidance (least-privilege at the workflow level).

## Checklist

- [x] No tests required (CI-only change)
- [x] No security vulnerabilities — this *adds* a security control
- [x] No docs update needed

.github/workflows/test.yml
jisung-02: Added this as a convention to CLAUDE.md so it's enforced going forward.

junho226 (approved):

PR #92, opened 2026-05-13
ci: use shared-ci dockerhub-login wrapper

choongbum:
## Summary

Replace the inline `docker/login-action@v3` step in `docker.yml` with the shared composite action `alpacax/shared-ci/common/dockerhub-login@v1`.

The wrapper hardcodes `docker.io` as registry and `alpacax` as username (both overridable via inputs), so callers only need to pass the push token. This reduces caller surface from 3 inputs to 1, and centralizes future updates (docker/login-action version bumps, registry migrations, username changes) in a single place.

### `.github/workflows/docker.yml`
- Replaced the inline `docker/login-action@v3` step with `alpacax/shared-ci/common/dockerhub-login@v1`.
- Migrated credentials from the legacy shared `DOCKER_USERNAME` / `DOCKER_PASSWORD` pair to the org-level `DOCKERHUB_PUSH_TOKEN` secret (backed by an AlpacaX org access token). This narrows blast radius if the token is ever compromised.
- The workflow no longer references `docker/login-action@v3` directly; the wrapper internally uses v4.

## Checklist

- [x] `DOCKERHUB_PUSH_TOKEN` secret available at org or repo level
- [ ] CI build job passes on this branch
- [ ] Image push works on next tag/main run

## Test plan

- [ ] PR Build Check passes (or whatever CI runs on PRs in this repo)
- [ ] After merge, the next `Docker` workflow run on `main` succeeds:
  - Log in to Docker Hub step uses the wrapper path
  - Build and push step succeeds
  - Image is published

## Notes

- The wrapper default `docker.io` / `alpacax` matches AlpacaX's existing Docker Hub setup. If this repo were using a different registry or username, the wrapper supports overriding via `with: registry: ...` / `with: username: ...`, but no override is expected here.
- This is part of an org-wide migration introduced by `alpacax/docs#242`. Same pattern will be applied to other Next.js / service repos in follow-up PRs.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -29,10 +29,9 @@ jobs:
         uses: docker/setup-buildx-action@v3
 
       - name: Log in to Docker Hub
-        uses: docker/login-action@v3
+        uses: alpacax/shared-ci/common/dockerhub-login@v1
         with:
-          username: ${{ secrets.DOCKER_USERNAME }}
-          password: ${{ secrets.DOCKER_PASSWORD }}
+          token: ${{ secrets.DOCKERHUB_PUSH_TOKEN }}
 
       - name: Extract metadata (tags, labels)
         id: meta


.github/workflows/docker.yml
```diff
@@ -29,10 +29,9 @@ jobs:
         uses: docker/setup-buildx-action@v3
 
       - name: Log in to Docker Hub
-        uses: docker/login-action@v3
+        uses: alpacax/shared-ci/common/dockerhub-login@v1
```
junho226, line 32: Warning (org-policy alignment): `alpacax/shared-ci/common/dockerhub-login@v1` is pinned to a mutable major tag. Since the action lives in `alpacax/shared-ci` it's inside our trust boundary, but GitHub's Actions hardening guidance recommends SHA-pinning for any action that lives in a different repo, because a compromised tag would still execute in our CI with the Dockerhub push token.

If other repos in the `alpacax/docs#242` migration also use `@v1`, please stay consistent with them — but if there's an org convention to SHA-pin third-party actions, this should follow it. Non-blocking; just flagging for org-level consistency.

junho226 (approved):
## Summary

Replaces the inline `docker/login-action@v3` step in `.github/workflows/docker.yml` with the shared composite action `alpacax/shared-ci/common/dockerhub-login@v1`, and migrates credentials from `DOCKER_USERNAME`/`DOCKER_PASSWORD` to a single org-level `DOCKERHUB_PUSH_TOKEN`. Part of the org-wide migration introduced in `alpacax/docs#242`.

## Verdict

**APPROVED** — small, single-responsibility CI consolidation with clear follow-up.

## Highlights

- Caller input surface shrinks from 3 inputs to 1 (token only); future updates to the underlying `docker/login-action` version, username, or registry are centralized in the wrapper.
- Switching to an org-level access token narrows credential blast radius and enables one-place rotate/revoke.
- Tightly scoped change (1 file, ±5 lines) — minimal regression risk.
- PR description clearly links to the org-wide migration (`alpacax/docs#242`) and identifies follow-up work.

## Warnings

1. `alpacax/shared-ci/common/dockerhub-login@v1` is pinned to a mutable major tag. Since this is an org-owned action it sits inside the trust boundary, but GitHub's Actions hardening guidance recommends SHA-pinning for *any* action that lives in a different repo. Match whatever convention the rest of the org uses (if other repos in the `docs#242` migration also pin to `@v1`, stay consistent; if they SHA-pin, switch).

## Suggestions

1. After this lands across the org-wide migration, remove the now-unused `DOCKER_USERNAME` / `DOCKER_PASSWORD` repo/org secrets to shrink secret surface. Worth tracking in the migration's parent issue.
2. Because Dockerhub login can't be dry-run in PR CI, leave a short comment on this PR (or the parent migration issue) after the first post-merge `Docker` workflow on `main` succeeds, so subsequent repo migrators have a known-good reference run.

## Checklist

- [x] Code follows stack/CI conventions
- [x] No security regressions (in fact, narrows credential scope)
- [x] PR body provides org-wide context
- [ ] Build verification deferred to post-merge (acceptable for this change)

jisung-02 (approved):

PR #98, opened 2026-05-14
simplify: remove unused websockets dependency

jisung-02:
## Summary

- `websockets` was listed as a runtime dependency but never imported anywhere in the codebase
- Its only appearance was a hardcoded version string inside the `MCP_USER_AGENT` constant in `utils/common.py`, which has been cleaned up
- `uv sync` confirms the package is uninstalled with no breakage (`Uninstalled 1 package: websockets==15.0.1`)

## Changes

- `pyproject.toml`: remove `websockets>=15.0.1` from `dependencies`
- `utils/common.py`: remove stale `websockets/15.0.1` token from `MCP_USER_AGENT` string
- `uv.lock`: updated automatically (57 packages, down from 58)

## Verification

```
$ uv sync
Resolved 57 packages in 19ms
Uninstalled 1 package in 6ms
 - websockets==15.0.1

$ uv lock --check
Resolved 57 packages in 4ms
```

## Test plan

- [x] `uv sync` exits cleanly and removes `websockets`
- [x] `uv lock --check` confirms lock file is consistent
- [x] No source file imports `websockets` (confirmed via `grep -r "import websockets"`)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -42,7 +42,6 @@ classifiers = [
 dependencies = [
     "mcp[cli]>=1.9.4",
     "httpx>=0.25.0",
-    "websockets>=15.0.1",
     "PyJWT[crypto]>=2.8.0",
 ]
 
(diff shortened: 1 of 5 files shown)


jisung-02:
Closing as requested.

PR #121, opened 2026-06-18
fix: standardize http_client error-envelope unwrapping to prevent false success

jisung-02:
### Background

`utils/http_client` does not raise on 4xx/5xx; it returns an error envelope dict (`{'error', 'status_code', 'message'}`). Many tools wrapped that return value in `success_response(data=result)` without checking, so an upstream error masqueraded as `status: "success"`. Callers then mistake auth expiry, 404, or permission errors for a normal response.

### Changes

**1. Standardize single-call tools** — `add1c82`

Added an `unwrap_http_result` guard across the single-call paths of 10 modules: alert, approval, audit, cert, events, iam, metrics, package, security, system_info.

```python
err = unwrap_http_result(result, default_message='...', region=region, workspace=workspace)
if err:
    return err
```

If the value is an error envelope it becomes an error_response that preserves `status_code`; otherwise it passes through and is wrapped normally.

**2. Tests and shared fixture** — `ef39de8`

- Added an error-path test per module, collapsed to one parametrized case per module.
- Hoisted the shared `HTTP_ERROR_ENVELOPE` constant into `tests/conftest.py`.
- Removed duplicated inline envelope literals and unused constants.

**3. Two aggregation-path false successes** — `82ae550`

- `get_disk_usage`: a 4xx/5xx on the device-discovery call fell through to the empty-list fallback and returned "No disk devices found", masking the real error. It now surfaces the real status code and skips the follow-up metrics call.
- `get_top_servers`: the single-metric path passed a gathered error envelope straight to `success_response`. Now guarded.

Reviewed but intentionally unchanged, since they are not false successes: `get_server_metrics_summary`, `get_server_overview`, `get_disk_info`. They already represent errors per metric or per key, so no error data is lost.

### Tests

```
818 passed, ruff All checks passed!
```

The two failing tests `test_import_main` and `test_main_entry_point` are an environment artifact: the working directory is named `main/`, so `import main` resolves to the package instead of `main.py`. They are unrelated to this change and pass in CI where the directory is `alpacon-mcp/`.

### Impact

Read-only error-handling improvement. No change to success response shape or behavior. No security, data-loss, or race risk.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -8,6 +8,14 @@ from unittest.mock import MagicMock, patch
 
 import pytest
 
+# Canonical http_client error envelope (the shape utils/http_client returns on
+# 4xx/5xx). Shared across error-path tests so the envelope is defined once.
+HTTP_ERROR_ENVELOPE = {
+    'error': 'HTTP Error',
+    'status_code': 404,
+    'message': 'Not found',
+}
+
 
 @pytest.fixture(autouse=True)
 def mock_region_auto_detect():
--- a/tests/test_alert_tools.py
+++ b/tests/test_alert_tools.py
@@ -1,0 +1,218 @@
+"""Unit tests for alert management tools."""
+
+from unittest.mock import AsyncMock, patch
+
+import pytest
+
+from tests.conftest import HTTP_ERROR_ENVELOPE
+from tools.alert_tools import (
+    create_alert_rule,
+    delete_alert_rule,
+    get_alert,
+    list_alerts,
+    mute_alert,
+    update_alert_rule,
+)
+
+ALERT_ID = 'alert-1'
+RULE_ID = 'rule-1'
+
+
+@pytest.fixture
+def mock_http_client():
+    with patch('tools.alert_tools.http_client') as mock_client:
+        mock_client.get = AsyncMock()
+        mock_client.post = AsyncMock()
+        mock_client.patch = AsyncMock()
+        mock_client.delete = AsyncMock()
+        yield mock_client
+
+
+@pytest.fixture
+def mock_token_manager():
+    with patch('utils.common.token_manager') as mock_manager:
+        mock_manager.get_token.return_value = 'test-token'
+        yield mock_manager
+
+
+class TestListAlerts:
+    @pytest.mark.asyncio
+    async def test_list_success(self, mock_http_client, mock_token_manager):
+        mock_http_client.get.return_value = {'results': [], 'count': 0}
+
+        result = await list_alerts(workspace='testworkspace', region='ap1')
+
+        assert result['status'] == 'success'
+        mock_http_client.get.assert_called_once_with(
+            region='ap1',
+            workspace='testworkspace',
+            endpoint='/api/alerts/',
+            token='test-token',
+            params={},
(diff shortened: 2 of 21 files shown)


junho226 (dismissed):
## Summary

`utils/http_client` does not raise on 4xx/5xx; it returns an error-envelope dict (`{'error', 'status_code', 'message'}`). Many tools wrapped that value in `success_response(data=result)` without checking, so an upstream error masqueraded as `status: "success"`. This PR extends the existing `unwrap_http_result` helper (already used in server/token/webftp/command modules) to the remaining 10 single-call modules (alert, approval, audit, cert, events, iam, metrics, package, security, system_info), plus fixes two metrics aggregation paths (`get_disk_usage` device discovery, `get_top_servers` single-metric path). Per-module error-path tests and a shared `HTTP_ERROR_ENVELOPE` fixture are added.

### Highlights

- **Solid helper design**: `unwrap_http_result` has a clear docstring, preserves `status_code`, and returns `None` for non-envelope payloads so callers flow naturally into `success_response`.
- **`get_disk_usage` fix is correct**: the guard sits *before* the empty-list fallback ("No disk devices found"), so the real error is not masked. The test asserts the follow-up metrics call is skipped (`call_count == 1`).
- **Coverage is complete**: on this branch the only module without the guard is `health_tools.py`, which makes no upstream HTTP call (local health check) and therefore cannot produce an error envelope. Not a gap.
- **Test quality**: per-module error-path tests plus a shared fixture remove duplication. Verified locally — 61 related tests pass.
- Intentionally-unchanged aggregation paths (`get_server_metrics_summary`, etc.) are documented with rationale in the PR body.

### Suggestions (optional, non-blocking)

1. `utils/common.py` — the error-envelope check `isinstance(result, dict) and 'error' in result` could misclassify a legitimate API payload that has a top-level `error` key. This is the codebase-wide convention (not introduced here), so it is acceptable; a dedicated marker (e.g. `__http_error__`) could be considered in a future refactor.
2. `get_top_servers` multi-metric path — an http error envelope is stored raw in `combined_data[metric]` rather than the `{'error': ..., 'available': False}` shape used on the exception path. Error data is preserved (not a false success), so this is acceptable as-is, but normalizing the shape would simplify consumer-side branching.

### Verdict

**APPROVED** — clean, well-scoped, read-only error-handling improvement with thorough test coverage. No critical issues or warnings; only two optional suggestions.

**Checklist**
- [x] Code follows stack conventions (consistent reuse of `unwrap_http_result`)
- [x] Tests included and passing (per-module error-path tests; verified locally)
- [x] No security vulnerabilities (read-only error-handling change)
- [x] Documentation updated if needed (intent and rationale in PR body)

jisung-02:
Thanks for both suggestions.

1. unwrap_http_result error-key heuristic — agreed it's the codebase-wide convention and not introduced here, so leaving it as-is. A dedicated __http_error__ marker is worth a separate refactor.
2. get_top_servers multi-metric shape — good catch, applied. The http error envelope is now normalized to the same {'error': ..., 'available': False} shape as the exception path before going into combined_data, so consumers branch on one shape.

junho226 (approved):

PR #122, opened 2026-06-23
feat: translate WorkSession gate errors for MCP agents

jisung-02:
## Summary

Translates alpacon-server's WorkSession gate errors (ADR 0014, zero standing privilege) into agent-actionable MCP results, and resolves an `ALPACON_WORK_SESSION` env fallback for the session id. The server already enforces the gate for interactive/OAuth callers; this PR is purely the client-side translation + env resolution (mirrors alpacon-cli's `worksession_error.go` + `resolve.go`). No server changes.

## What changed

- Gate error translation (`utils/common.py`): `unwrap_http_result` now parses the server's `{"code": "<gate_code>"}` 400 body and maps it: `work_session_not_active` becomes `status="pending_approval"` (human must approve out-of-band, then retry); the 6 other gate codes become `status="error"` with a static `code` + `next_action` telling the agent how to get inside a valid session. Centralized in `unwrap_http_result`, so every tool using it is covered with zero per-tool changes.
- Env fallback (`resolve_work_session_id`): `work_session_id` arg > `ALPACON_WORK_SESSION` env > `None`, both paths whitespace-stripped.
- Command tools (`tools/command_tools.py`): route single-server, multi-server (parallel + sequential) error paths through `unwrap_http_result` so gate codes are translated on every command execution path, and resolve the session id via `resolve_work_session_id` before injecting it into the payload.
- WebFTP tools (`tools/webftp_tools.py`): resolve the session id via `resolve_work_session_id` at all session-id injection sites, so file-transfer paths get the same env fallback and gate translation.
- Docs: CLAUDE.md documents the gate behavior and the env var.

## Security properties

- `next_action` strings come only from a static dict; no server-supplied text is echoed to the agent.
- Gate `code` is allowlist-gated by a 7-member frozenset; unknown codes fall back to a generic error (server keeps enforcing regardless).
- No raw error envelope leaks through (test-covered).

## Testing

`uv run pytest tests/test_common.py tests/test_command_tools.py tests/test_webftp_tools.py -q` passes. Coverage:
- `test_common.py` (21 tests): all 7 gate codes, pending vs. error mapping, non-gate/non-JSON/missing-body fallbacks, and env resolution (explicit > env > none, whitespace-only handling).
- `test_command_tools.py`: gate translation on single- and multi-server execution paths, no raw-envelope leak, and env-fallback session resolution.
- `test_webftp_tools.py`: env-fallback session resolution on file-transfer paths.


tools/command_tools.py
```diff
@@ -245,11 +246,16 @@ async def execute_command(
...
     if isinstance(exec_data, dict) and 'error' in exec_data:
-        return error_response(
-            f'Command execution failed: {exec_data.get("error", "Unknown error")}',
-            workspace=workspace,
-            region=region,
-            details=exec_data,
+        # unwrap_http_result returns non-None whenever 'error' is in the dict
+        return cast(
+            dict[str, Any],
+            unwrap_http_result(
```
junho226, line 252: Suggestion (informational): the message format for non-gate command errors changes here. Previously this returned `f'Command execution failed: {exec_data.get("error")}'` plus `details=exec_data`; now `unwrap_http_result` returns only `result.get('message', 'Command execution failed')` and drops `details`. This is a deliberate, test-covered security improvement (no raw envelope leak), but it reduces inline debug info for non-gate errors. Fine as long as the upstream envelope always carries a `message` field.

tools/command_tools.py
```diff
@@ -416,7 +422,13 @@ async def _submit_one(sid: str) -> dict[str, Any] | list[Any]:
                 }
                 failed_count += 1
             elif isinstance(result, dict) and 'error' in result:
-                deploy_results[sid] = {'status': 'error', **result}
+                deploy_results[sid] = unwrap_http_result(
```
junho226, line 426: Suggestion (nit): the single-server path has an explanatory comment about the non-None invariant of `unwrap_http_result`, but the two multi-server branches don't. Consider adding the same comment here for consistency.

junho226 (dismissed):
## PR Review: #122 - feat: translate WorkSession gate errors for MCP agents

**Stack**: python | **Changes**: +435 -25 in 7 files

---

### Summary
Translates alpacon-server's WorkSession gate errors (ADR 0014, zero standing privilege) into structured, agent-actionable MCP results, and resolves an `ALPACON_WORK_SESSION` env fallback for the session id. The translation is centralized in `unwrap_http_result`, so every tool using it is covered with zero per-tool changes. Pure client-side work that mirrors alpacon-cli's `worksession_error.go` + `resolve.go`; no server changes.

**Verification performed:**
- `utils/common.py` — gate extraction runs only on dicts containing `error`. `_extract_work_session_gate_code` JSON-parses `response` and returns a code only if it is in the 7-member allowlist (frozenset); missing/non-str/non-JSON/non-dict bodies all fall back to None safely.
- `command_tools.py` — single-server guard `isinstance(exec_data, dict) and 'error' in exec_data` and multi-server (parallel/sequential) guard `isinstance(result, dict) and 'error' in result` both guarantee `unwrap_http_result` returns non-None, so the `cast`/assignment is safe. No None-assignment bug.
- `pending_approval_response`'s `WORK_SESSION_PENDING` category is registered in `_NEXT_ACTION_BY_CATEGORY`, so it does not fall back to the sudo default and gets the correct work-session `next_action`.
- Both `error_response` and `pending_approval_response` accept `**kwargs`, so the new `code`/`next_action`/`requires_human_approval` fields flow through correctly. `pending_approval_response` applies its security fields (flags/category) after kwargs so they cannot be overridden.
- `resolve_work_session_id`: explicit > env > None, both stripped. Correct.

### Critical Issues (0)
None.

### Warnings (0)
None.

### Suggestions (2)
See inline comments.

### Highlights
- Centralizing gate translation in `unwrap_http_result` — DRY, zero per-tool changes. Excellent design.
- Strong security properties: `next_action` comes only from a static dict, `code` is allowlist-gated by a 7-member frozenset, and no raw envelope leaks through (test-covered). The server keeps enforcing regardless.
- Comprehensive test coverage: all 7 gate codes, pending vs. error mapping, non-gate/non-JSON/missing-body fallbacks, env precedence (explicit > env > none) and whitespace handling, single/multi-server execution paths, and the WebFTP path.
- Walrus operator (`if ws_id := resolve_work_session_id(...)`) cleanly handles the 7 WebFTP injection sites.
- Mirroring alpacon-cli keeps behavior consistent across the platform.

---

**Verdict**: APPROVED

**Checklist**:
- [x] Tests included (comprehensive; not independently run here since the working tree is on main — relying on the PR's reported pass and the clean Copilot review)
- [x] No security vulnerabilities (allowlist + static next_action + no envelope leak)
- [x] Documentation updated (CLAUDE.md documents the gate behavior and the env var)

tools/command_tools.py
jisung-02: Confirmed the gating assumption holds. Every error envelope produced by `utils/http_client.py` carries a `message` field — all 7 paths: 401/MFA (l.235), 4xx client error (l.395, the path non-gate command errors take), Timeout (l.414), Request Error (l.430), Unexpected Error (l.436), Max retries exceeded (l.443), and Request Exception (l.514). On top of that, `unwrap_http_result` falls back to `default_message` (`result.get('message', default_message)`), and the command path passes "Command execution failed", so even a message-less envelope degrades to the previous wording rather than breaking. No change needed.

tools/command_tools.py
jisung-02: Done — added the same `# unwrap_http_result returns non-None whenever 'error' is in the dict` comment to both multi-server branches (parallel and sequential) for consistency. Pushed.

junho226 (approved):

PR #128, opened 2026-07-07
fix(tools): stop wrapping HTTP failures as success responses

### LINKED ISSUE #127: Fix server/webhook tools wrapping HTTP failures as successful responses
## Background

  Some MCP tools wrap failed `http_client` responses with `success_response()` without checking the result first.

  `http_client` can return 4xx/5xx responses as a dict instead of raising an exception:

  ```python
  {
      "error": "HTTP Error",
      "status_code": 403,
      "message": "...",
      "response": "..."
  }

  The affected tools pass this value directly into success_response(), which can produce a final MCP response like this:

  {
      "status": "success",
      "data": {
          "error": "HTTP Error",
          "status_code": 403,
          "message": "..."
      }
  }
```
  In other words, the upstream API call failed, but the MCP client may interpret the tool call as successful.

  ## Impact

  This is especially risky because the affected tools include destructive operations such as reboot_system, shutdown_system, restart_agent, shutdown_agent, and upgrade_system.

  For example, if the upstream API fails due to missing permissions, a Work Session gate, 404, or 403, the agent may still interpret the operation as successfully completed.

  ## Scope

  tools/server_tools.py

  - list_server_notes
  - create_server_note
  - restart_agent
  - shutdown_agent
  - upgrade_agent
  - update_information
  - upgrade_system
  - reboot_system
  - shutdown_system

  tools/webhook_tools.py

  - list_event_subscriptions
  - create_event_subscription
  - delete_event_subscription
  - list_webhooks
  - create_webhook
  - update_webhook
  - delete_webhook

  Excluded:

  - get_webhook: already uses unwrap_http_result()
  - list_servers, get_server: already have manual error guards
  - get_disk_info: aggregator behavior should be handled separately

  ## Proposed Fix

  Apply the existing standard pattern after each http_client call:
```python
  err = unwrap_http_result(
      result,
      default_message="Failed to ...",
      ...
  )
  if err:
      return err

  return success_response(data=result, ...)
```
  ## Tests

  Add regression coverage for:

  - http_client returning {"error": "HTTP Error", "status_code": 403, "message": "..."}
  - Final MCP response is status: error
  - status_code is preserved in the final error response
  - Normal success responses still return status: success
  - At least one destructive server tool and one webhook tool

  ## Completion Criteria

  - All 16 affected tools pass through unwrap_http_result() or an equivalent error guard
  - 4xx/5xx error dicts are no longer wrapped as successful responses
  - Existing success-path behavior is preserved
  - Related tests are added and passing

jisung-02:
## Background

`utils/http_client` does not raise on upstream 4xx/5xx. Instead it returns an error envelope dict shaped like `{'error':..., 'status_code':..., 'message':...}`. Several tools passed that return value straight into `success_response(data=result, ...)` without inspecting it, so a failed upstream call surfaced to the MCP client as `status="success"`. This also masked WorkSession gate denials (403), hiding cases that actually need out-of-band human approval behind a fake success.

Most tools in the repo already route their result through `unwrap_http_result()` (which converts an error envelope into an `error_response`, or a WorkSession gate response, and otherwise returns `None`). Only a few tools were missing that guard; this PR closes those gaps.

## Changes

Add the missing `unwrap_http_result()` check to the server, webhook, and system-info tools so HTTP failures are no longer wrapped as success.

- `tools/server_tools.py`: guard `list_server_notes`, `create_server_note`, and the seven agent/system actions (`restart_agent`, `shutdown_agent`, `upgrade_agent`, `update_information`, `upgrade_system`, `reboot_system`, `shutdown_system`). The seven actions shared an identical `POST /actions/` call plus unwrap, so they were factored into a `_server_action()` private helper.
- `tools/webhook_tools.py`: guard the eight event-subscription and webhook tools (`list/create/delete_event_subscription`, `list/create/update/delete_webhook`).
- `tools/system_info_tools.py`: `get_disk_info` fetches `/proc/disks` and `/proc/partitions` concurrently via `asyncio.gather` and previously only checked for raised exceptions. Each result now hard-fails when it is an exception (`error_response`) or an error envelope (`unwrap_http_result`). Disks and partitions come from the same server and token, so a half-populated payload is not a useful partial success.

## Safety

This change improves authz visibility. Previously a WorkSession gate denial (403 with `work_session_required` and similar codes) also returned `status="success"`, making a denial look like a success. Those gate codes now route through `work_session_gate_response`, matching the gating semantics already used by the other tools. `unwrap_http_result` only forwards `message`, `status_code`, and the fixed strings for recognized gate codes, never the raw upstream response body, so surfacing the failure does not leak sensitive upstream data. Tokens are sent via the `Authorization` header and never appear in error messages.

## Tests

- Add regression tests asserting the error envelope is not wrapped as success for each tool (`test_agent_actions.py`, `test_server_tools.py`, `test_webhook_tools.py`, `test_system_info_tools.py`).
- For `get_disk_info`, assert that both the raised-exception path and the error-envelope path return `status="error"`, and that the envelope path preserves `status_code`. Two existing tests that pinned the old partial-success behavior (`test_system_info_tools.py`, `test_system_tools.py`) were updated to the new behavior.
- Centralize the repeated error envelope and token-manager mock into shared `tests/conftest.py` helpers (`HTTP_ERROR_ENVELOPE`, `mock_token_manager`).
- Full suite: 882 passed.

## Checklist

- [x] Every http_client call site checks for an error envelope before `success_response` (audited across all tools modules; this branch handles the last gap, `get_disk_info`)
- [x] Passed a six-stage review loop (correctness, performance, security, clean code, over-engineering, AlpacaX conventions)
- [x] Regression tests added; full suite green

Closes #127

### DIFF
--- a/tools/system_info_tools.py
+++ b/tools/system_info_tools.py
@@ -395,6 +395,10 @@ async def get_disk_info(
             workspace=workspace,
         )
         if err:
+            # Work Session gate responses already carry a specific message; only
+            # plain upstream errors need the disks-vs-partitions label restored.
+            if err.get('status') == 'error' and 'code' not in err:
+                err['message'] = f'{default_message}: {err["message"]}'
             return err
 
     disk_info = {
--- a/tools/system_info_tools.py
+++ b/tools/system_info_tools.py
@@ -3,7 +3,7 @@
 import asyncio
 from typing import Any
 
-from utils.common import success_response, unwrap_http_result
+from utils.common import error_response, success_response, unwrap_http_result
 from utils.decorators import mcp_tool_handler
 from utils.http_client import http_client
 from utils.tool_annotations import READ_ONLY
@@ -376,15 +376,31 @@ async def get_disk_info(
         if isinstance(r, BaseException) and not isinstance(r, Exception):
             raise r
 
-    # Prepare response
+    for result, default_message in (
+        (disks_result, 'Failed to get disk info'),
+        (partitions_result, 'Failed to get partition info'),
+    ):
+        if isinstance(result, Exception):
+            return error_response(
+                f'{default_message}: {result}',
+                server_id=server_id,
+                region=region,
+                workspace=workspace,
+            )
+        err = unwrap_http_result(
+            result,
+            default_message=default_message,
+            server_id=server_id,
+            region=region,
+            workspace=workspace,
+        )
+        if err:
+            return err
+
     disk_info = {
         'server_id': server_id,
-        'disks': disks_result
-        if not isinstance(disks_result, Exception)
-        else {'error': str(disks_result)},
-        'partitions': partitions_result
-        if not isinstance(partitions_result, Exception)
-        else {'error': str(partitions_result)},
+        'disks': disks_result,
+        'partitions': partitions_result,
         'region': region,
         'workspace': workspace,
     }
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -17,6 +17,14 @@ HTTP_ERROR_ENVELOPE = {
 }
 
 
+@pytest.fixture
+def mock_token_manager():
+    """Mock token manager so tests never read a real ~/.alpacon-mcp/token.json."""
+    with patch('utils.common.token_manager') as mock_manager:
+        mock_manager.get_token.return_value = 'test-token'
+        yield mock_manager
+
+
 @pytest.fixture(autouse=True)
 def mock_region_auto_detect():
     """Prevent _resolve_region from accessing real token.json in all tests.
--- a/tests/test_agent_actions.py
+++ b/tests/test_agent_actions.py
@@ -4,6 +4,7 @@ from unittest.mock import AsyncMock, patch
 
 import pytest
 
+from tests.conftest import HTTP_ERROR_ENVELOPE
 from tools.server_tools import (
     reboot_system,
     restart_agent,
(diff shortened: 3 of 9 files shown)


tools/system_info_tools.py
```diff
@@ -376,15 +376,31 @@ async def get_disk_info(
...
+        (partitions_result, 'Failed to get partition info'),
+    ):
+        if isinstance(result, Exception):
+            return error_response(
+                f'{default_message}: {result}',
+                server_id=server_id,
+                region=region,
+                workspace=workspace,
+            )
+        err = unwrap_http_result(
```
junho226, line 390: Suggestion: on the error-envelope path, `unwrap_http_result()` returns `result.get('message', default_message)`. Since the upstream envelope usually *has* a `message`, the `default_message` ("Failed to get partition info" vs "Failed to get disk info") is discarded here — so the caller can't tell whether `disks` or `partitions` failed. The `isinstance(result, Exception)` branch just above preserves that distinction via `f'{default_message}: {result}'`. Consider carrying the sub-call label into the envelope-path message too for symmetric debuggability.

Minor nit — fine to merge as-is.

junho226 (dismissed):
## Summary

Closes the remaining gaps where `http_client`'s 4xx/5xx error envelope (`{'error', 'status_code', 'message'}`) was passed straight into `success_response()`, so failed upstream calls surfaced as `status="success"`. It routes the last unguarded call sites in `server_tools`, `webhook_tools`, and `system_info_tools` through the repo-standard `unwrap_http_result()`, which also translates WorkSession gate denials (403) into gate/pending-approval responses instead of masking them as success. The seven identical agent/system actions are factored into a `_server_action()` helper.

**Change type**: bug fix (correctness + authz visibility). Well-scoped, defensive, thoroughly tested.

### Critical issues (0)
None.

### Warnings (0)
None.

### Suggestions (1)
1. **`get_disk_info` error granularity** — on the error-envelope path, `unwrap_http_result()` returns the upstream `message` and discards `default_message`, so the caller loses the disks-vs-partitions distinction (the raised-exception path preserves it). Minor debuggability nit, acceptable to merge as-is. See inline comment.

### Highlights
- **Correct root-cause fix**: aligns with the established `unwrap_http_result()` convention; imports already present in all three files (no `NameError` risk), and `error_response` correctly added to the `system_info_tools` import.
- **Security-positive**: WorkSession gate denials (403) previously looked like success — a genuine authz-visibility bug. They now route through `work_session_gate_response`, and no raw upstream body is leaked (only `message`/`status_code`/recognized gate codes).
- **Good DRY refactor**: `_server_action()` collapses seven near-identical POST-and-unwrap bodies (keyword-only args, defined before first use).
- **`get_disk_info` fail-fast** is correct: the `BaseException` re-raise (e.g. `CancelledError`) is preserved before the new Exception/envelope checks; partial-success masking is removed intentionally.
- **Strong test coverage**: parametrized regression tests for all touched tools; shared `HTTP_ERROR_ENVELOPE` + `mock_token_manager` centralized in `conftest.py`; the two behavior-pinning tests were correctly updated to the new contract rather than left stale.

### Verdict: APPROVED

- [x] Code follows stack conventions (matches existing `unwrap_http_result` pattern)
- [x] Tests included and passing (882 passed; parametrized regression coverage added)
- [x] No security vulnerabilities (improves authz visibility; no data leakage)
- [x] Documentation adequate (minor: PR body says "eight" webhook tools but modifies seven — `get_webhook` was already guarded)

_Reviewed with `/alpacax:review-pr`._

tools/system_info_tools.py
jisung-02: Fixed in 5adc750: prefix the disks/partitions default_message onto the plain error_response path (skipping Work Session gate responses, which already carry a specific message), so both branches now report which sub-call failed.

junho226 (approved):

PR #131, opened 2026-07-13
refactor: quarantine Any into a typed boundary and make mypy strict a CI gate

jisung-02:
## Background

The codebase handled most API responses and tool return values as `dict[str, Any]`, so mypy verified essentially nothing. The CI mypy step was also gated with `continue-on-error: true`, so type errors passed silently. This PR removes `Any` across the codebase, quarantines it into a single boundary module, and promotes mypy strict to a CI gate that actually fails.

## What changed

Adds a new module `utils/api_types.py` that confines the unavoidable `Any` to this one file. Untrusted JSON crossing the API boundary is expressed only through explicit boundary aliases such as `ApiPayload`, `JwtClaims`, `AsgiMessage`, and `JsonObject`, and every other module imports these aliases instead of using `Any` directly. All `Any` used as an actual type annotation now lives in this module; the 3 remaining `Any` tokens in other files are all text inside comments.

Formalizes the tool response envelope with TypedDicts. Defines `SuccessResponse`, `ErrorResponse`, `ValidationErrorResponse`, `PendingApprovalResponse`, and their union `ToolResponse`. The context keys each tool attaches to a response were enumerated via an AST scan of every keyword argument passed to the response builders and collected into the `ResponseContext` TypedDict. The builders accept `**kwargs: Unpack[ResponseContext]`, so an invalid key is caught at type-check time.

Narrows raw JSON parsing in the HTTP client to a single `_parse_json_body()` boundary function, and aligns every method to return `ApiResult` (either a valid payload or the standardized error envelope). Tool code branches on errors via the `is_api_error()` TypeGuard.

Applies PEP 695 generics and ParamSpec to the decorators so that the original tool signature is preserved after `mcp_tool_handler` and friends wrap it. While doing this we found a bug where FastMCP's `func_metadata()` did not filter out the token-injection `**kwargs`, exposing a required `kwargs: string` field in every tool's input schema. Fixed by stripping the VAR_KEYWORD parameter from the exposed signature.

Sets `[tool.mypy] strict = true` in `pyproject.toml` and removes `continue-on-error` from the CI mypy step. Type errors now fail CI.

## Safety

The guiding principle was no runtime behavior change. The only observable change is the FastMCP schema bug fix above: the erroneously exposed `kwargs` field disappears from tool input schemas. This is pinned by the regression test `tests/test_tool_schemas.py`.

## Testing

Locally: ruff check passes, ruff format verified (83 files), mypy strict reports no issues across 39 source files, pytest 868 passed. Added unit tests for `api_types.py` and a tool-schema regression test.

## Checklist

- [x] mypy strict passes (39 files)
- [x] full test suite passes (868 passed)
- [x] ruff lint/format passes
- [x] no runtime behavior change (except the FastMCP schema bug fix)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -51,8 +51,7 @@ jobs:
     - name: Type check with mypy
       run: |
         uv pip install mypy
-        uv run --no-sync mypy --ignore-missing-imports --no-strict-optional .
-      continue-on-error: true
+        uv run --no-sync mypy
 
     - name: Test with pytest
       run: |
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -128,6 +128,7 @@ This MCP server provides a **pure HTTP API bridge** to Alpacon's infrastructure 
 - `cert_tools.py`: Certificate authority, signing requests, and certificate management
 
 **Utilities** (all in `utils/` directory):
+- `api_types.py`: Shared TypedDicts and type aliases for tool responses and API-boundary payloads (the sole sanctioned use of `Any`)
 - `http_client.py`: Async HTTP client for Alpacon API
 - `token_manager.py`: Secure token storage and management
 - `decorators.py`: MCP tool decorators (token validation, input validation, error handling, logging)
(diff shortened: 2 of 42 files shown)


jisung-02:
Close this PR due to too large modification range

PR #134, opened 2026-07-15
fix(commands): detect exec completion via handled_at, not finished_at

jisung-02:
## Background

Investigating the Slack report of unstable exec (server listing intermittently failing, and command execution timing out without returning a result even with a Work Session attached), the exec timeout turned out to be a 100% reproducible structural bug, not an intermittent one.

`execute_command`'s polling loop detected completion by reading a `finished_at` field, which the Alpacon Command API never returns. The real completion marker is `handled_at`, and completion status is computed by alpacon-server's `compute_status` (`events/models.py`) as `success`/`failed`. Since `finished_at` is always absent, a command that finished instantly was never recognized as complete, so the loop kept polling until its deadline and the client timed out first.

Reproduced live against the web-editor server on dev: the command (`whoami`) finished server-side every time in `response_delay` 0.9s / `elapsed_time` 0.009s with `status: success`, yet the MCP tool call never received the result and timed out. By contrast `execute_command_multi_server`, which only submits without polling, returned normally. That contrast confirms the submit path is fine and the wait (polling) path is the cause.

This `finished_at` check has never matched the real API since the initial implementation on 2025-09-24, so this was always broken from the start rather than a recent regression.

## Changes

`tools/command_tools.py` polling loop:

- Detect completion via `handled_at is not None`. Per `compute_status`, `handled_at` being set implies status is `success`/`failed`, so it is the single accurate completion signal.
- `awaiting_approval` (HITL verification gate) now returns a structured `pending_approval` result instead of continuing to poll. The command has not run and a human must approve out-of-band (ADR 0015), so there is no reason to burn the timeout window.
- `denied`/`rejected` join the existing `stuck`/`error` fast-fail branch. All are terminal states where the command does not run.

`utils/common.py`:

- Register `COMMAND_AWAITING_APPROVAL` in `_NEXT_ACTION_BY_CATEGORY`. It previously fell back to the generic default text; explicit registration is this codebase's convention, matching every other category.

`list_commands` description fix:

- The description pointed a "Related" reference at a `get_command_result` tool that does not exist. Since that is exactly the path for retrieving a result after an exec timeout, an agent following it would call a nonexistent tool and fail. It now points to the real recovery path (retrieve a timed-out command's result via `list_commands`) and the real sibling (`execute_command`).

## Safety

The `awaiting_approval` `pending_approval` response does not forward the raw poll payload (`data=result`). The Command API payload carries AI risk internals such as `risk_score`/`tier2_reasoning`, which ADR 0015 forbids disclosing to clients beyond the category and next_action. This is consistent with the sudo-denial path, which also discloses only the category.

## Tests

Nine existing fixtures that pinned the buggy behavior were switched to the real API shape (`handled_at` + `success`/`failed`, removing the fictitious `finished_at`/`completed`). New tests: completion detection for a failed command, a parametrized test over the four terminal statuses (`stuck`/`denied`/`rejected`/`error`), and the `awaiting_approval` pending path with a `next_action` assertion confirming the registered category.

Before the fix, the new fixtures hung the tests (the polling loop failing to detect completion, matching the live symptom); after the fix, `test_command_tools.py` passes 41 tests and the full suite passes 885. Two pre-existing failures (`test_import_main`, `test_main_entry_point`) were confirmed via `git stash` to fail identically on a clean tree, so they are unrelated to this change.

## Remaining limitation (out of scope for this PR)

Commands taking longer than 60s can still fail to return a result after this fix, due to two layers: (1) the alpacon-mcp ingress has no ALB `idle_timeout` setting, so the connection is dropped at the default 60s while nothing streams during polling (needs a `load-balancer-attributes: idle_timeout.timeout_seconds` annotation like the alpacon chart's WebSocket ingress, in the gitops repo); (2) the MCP client's own tool timeout (~60-80s observed) is outside our control. For long commands the correct pattern is submit-then-retrieve via `list_commands`, and this PR also corrects that recovery path in the documentation.

## Checklist

- [x] Completion detected via a real API field (`handled_at`), dropping the nonexistent `finished_at` dependency
- [x] Terminal states (`denied`/`rejected`/`awaiting_approval`) handled explicitly, removing wasted polling
- [x] `COMMAND_AWAITING_APPROVAL` category registered and pinned with a `next_action` test
- [x] Removed the phantom tool reference in `list_commands`
- [x] Bug-pinning tests updated to the real API shape; regression tests added
- [x] Confirmed the two pre-existing failures are unrelated to this branch

### DIFF
--- a/tools/command_tools.py
+++ b/tools/command_tools.py
@@ -158,7 +158,7 @@ async def _get_command_result(
 
 
 @mcp_tool_handler(
-    description='List recent command execution history with status, output, and timestamps. Filterable by server ID. When to use: reviewing past commands or checking what has been run on a server. Related: get_command_result (get full output of a specific command).',
+    description='List recent command execution history with status, output, and timestamps. Filterable by server ID. When to use: reviewing past commands, or retrieving the result of a command whose execute_command call timed out. Related: execute_command (run a command and wait).',
     annotations=READ_ONLY,
     meta={'anthropic/searchHint': 'command history list recent'},
 )
@@ -309,7 +309,9 @@ async def execute_command(
         if isinstance(result, dict):
             status = result.get('status', '')
 
-            if result.get('finished_at') is not None:
+            # handled_at is set once the agent reports the result (status then
+            # becomes 'success'/'failed'); the Command API has no 'finished_at'.
+            if result.get('handled_at') is not None:
                 response = success_response(
                     data=result,
                     command_id=command_id,
@@ -335,7 +337,25 @@ async def execute_command(
                         )
                 return response
 
-            if status in ('stuck', 'error'):
+            if status == 'awaiting_approval':
+                # HITL verification gate: a human must approve out-of-band
+                # (ADR 0015); polling would only burn the timeout window.
+                return pending_approval_response(
+                    'Command is awaiting human approval. A human must approve '
+                    'it out-of-band (Alpacon web console or Slack); then check '
+                    'the result via list_commands or re-run.',
+                    category='COMMAND_AWAITING_APPROVAL',
+                    command_id=command_id,
+                    server_id=server_id,
+                    command=command,
+                    region=region,
+                    workspace=workspace,
+                )
+
+            # Terminal non-approval statuses: the command will not produce a
+            # result (denied/rejected never run; stuck/error gave up), so fail
+            # fast instead of polling until timeout.
+            if status in ('denied', 'rejected', 'stuck', 'error'):
                 return error_response(
                     f'Command failed with status: {status}',
                     command_id=command_id,
--- a/tools/command_tools.py
+++ b/tools/command_tools.py
@@ -342,8 +342,10 @@ async def execute_command(
                 # (ADR 0015); polling would only burn the timeout window.
                 return pending_approval_response(
                     'Command is awaiting human approval. A human must approve '
-                    'it out-of-band (Alpacon web console or Slack); then check '
-                    'the result via list_commands or re-run.',
+                    'it out-of-band (Alpacon web console or Slack); it then runs '
+                    'automatically. Wait, then retrieve the result via '
+                    'list_commands. Do not re-run: a resubmission needs its own '
+                    'approval and may double-execute the command.',
                     category='COMMAND_AWAITING_APPROVAL',
                     command_id=command_id,
                     server_id=server_id,
@@ -353,9 +355,11 @@ async def execute_command(
                 )
 
             # Terminal non-approval statuses: the command will not produce a
-            # result (denied/rejected never run; stuck/error gave up), so fail
-            # fast instead of polling until timeout.
-            if status in ('denied', 'rejected', 'stuck', 'error'):
+            # result (denied/rejected never run; stuck gave up), so fail fast
+            # instead of polling until timeout. ('error' is omitted: the
+            # server's compute_status never emits it given the scheduled_at
+            # default, so it is unreachable here.)
+            if status in ('denied', 'rejected', 'stuck'):
                 return error_response(
                     f'Command failed with status: {status}',
                     command_id=command_id,
@@ -366,8 +370,19 @@ async def execute_command(
                     details=result,
                 )
 
-            # Command still in progress — reset deadline (within hard cap)
-            if status in ('running', 'acked'):
+            # Command still in progress — reset deadline (within hard cap) so a
+            # slow AI verification or delayed delivery does not time out a
+            # command that is still advancing. Covers both pre-execution states
+            # (queued/scheduled/delivered/verifying) and execution (running).
+            # 'acked' is intentionally absent: compute_status returns 'running'
+            # once acked_at is set, so it is never emitted.
+            if status in (
+                'running',
+                'verifying',
+                'delivered',
+                'queued',
+                'scheduled',
+            ):
                 deadline = min(
                     loop.time() + timeout,
                     hard_deadline,
--- a/utils/common.py
+++ b/utils/common.py
@@ -91,8 +91,10 @@ _NEXT_ACTION_BY_CATEGORY: dict[str, str] = {
     ),
     'COMMAND_AWAITING_APPROVAL': (
         'A human must approve this command out-of-band (Alpacon web console or '
-        'Slack). You cannot approve it yourself. Wait for approval, then check '
-        'the result via list_commands or re-run; do not repeatedly resubmit.'
+        'Slack); it then runs automatically. You cannot approve it yourself. '
+        'Wait for approval, then retrieve the result via list_commands. Do not '
+        're-run: a resubmission needs its own approval and may double-execute '
+        'the command.'
     ),
     'SUDO_PRESENCE_REQUIRED': (
         'A human must complete a fresh MFA step-up out-of-band, then retry. You '
--- a/utils/common.py
+++ b/utils/common.py
@@ -75,8 +75,9 @@ def success_response(data: Any = None, **kwargs) -> dict[str, Any]:
 # The human-resolvable next action differs by category: SUDO_APPROVAL_REQUIRED /
 # WORK_SESSION_PENDING need an out-of-band approval, while SUDO_PRESENCE_REQUIRED
 # is an MFA step-up and SUDO_NO_WORKSESSION_POLICY is a scope addition — none of
-# which the agent can perform itself. APPROVAL_DECISION_HUMAN_ONLY is a pure
-# explanation that approving/rejecting is human-only; there is nothing to retry.
+# which the agent can perform itself. COMMAND_AWAITING_APPROVAL is the
+# command-level analogue of WORK_SESSION_PENDING. APPROVAL_DECISION_HUMAN_ONLY is
+# a pure explanation that approving/rejecting is human-only; nothing to retry.
 _NEXT_ACTION_BY_CATEGORY: dict[str, str] = {
     'SUDO_APPROVAL_REQUIRED': (
         'A human must approve this out-of-band (Alpacon web console or Slack). '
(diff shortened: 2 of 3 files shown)


tools/command_tools.py
```diff
@@ -335,7 +337,25 @@ async def execute_command(
...
                 return response
 
-            if status in ('stuck', 'error'):
+            if status == 'awaiting_approval':
+                # HITL verification gate: a human must approve out-of-band
+                # (ADR 0015); polling would only burn the timeout window.
+                return pending_approval_response(
+                    'Command is awaiting human approval. A human must approve '
+                    'it out-of-band (Alpacon web console or Slack); then check '
+                    'the result via list_commands or re-run.',
```
junho226: Warning: This message advises "check the result via list_commands or re-run." For `awaiting_approval` the original command is pending delivery and will execute once a human approves — HITL approval sets `verification_status='approved'` and calls `command.execute()` (alpacon-server `events/services.py:223-238`), after which the same command reaches `handled_at`. "Re-run" therefore submits a duplicate command that needs its own separate approval and can double-execute an already-risk-flagged HITL command. The correct recovery is to wait and retrieve via `list_commands`. Unlike the sudo-denial categories (nothing delivered, re-run is the recovery), re-run is counterproductive here. Recommend dropping "or re-run". Non-blocking.

utils/common.py
```diff
@@ -88,6 +89,11 @@ def success_response(data: Any = None, **kwargs) -> dict[str, Any]:
         'or Slack) before it activates. You cannot approve it yourself. Wait for '
         'approval, then retry.'
     ),
+    'COMMAND_AWAITING_APPROVAL': (
+        'A human must approve this command out-of-band (Alpacon web console or '
+        'Slack). You cannot approve it yourself. Wait for approval, then check '
+        'the result via list_commands or re-run; do not repeatedly resubmit.'
```
junho226: Warning: Same "or re-run" issue in the registered `COMMAND_AWAITING_APPROVAL` next_action, and it's in tension with "do not repeatedly resubmit" in the same sentence. For `awaiting_approval` the command runs on human approval (alpacon-server `events/services.py:223-238`), so re-running duplicates it. Recommend: "Wait for approval, then retrieve the result via list_commands; do not resubmit." Non-blocking.

tools/command_tools.py
```diff
@@ -335,7 +337,25 @@ async def execute_command(
...
+                    server_id=server_id,
+                    command=command,
+                    region=region,
+                    workspace=workspace,
+                )
+
+            # Terminal non-approval statuses: the command will not produce a
+            # result (denied/rejected never run; stuck/error gave up), so fail
+            # fast instead of polling until timeout.
+            if status in ('denied', 'rejected', 'stuck', 'error'):
```
junho226: Suggestion: `'error'` here is a status the server never actually emits — `compute_status` marks the `return 'error'` branch unreachable (`# type: ignore[unreachable]`, alpacon-server `events/models.py:680`), since a Command always has a `scheduled_at` default. Harmless as defensive coding, but consider a comment or removing it for clarity. Also note (pre-existing, just below): only `status in ('running', 'acked')` resets the deadline, so in-progress states the server can return — `verifying`, `delivered`, `queued`, `scheduled` — poll against the original deadline, and `'acked'` is dead (compute_status returns `'running'`, never `'acked'`). A slow AI `verifying` phase could time the call out; acceptable given submit-then-retrieve, but worth documenting.

junho226 (dismissed):
## PR Review: #134 - fix(commands): detect exec completion via handled_at, not finished_at
**Author**: @jisung-02
**Branch**: fix/exec-command-completion-detection → main
**Stack**: python
**Changes**: +125 -28 in 3 files

### Summary
The PR fixes a 100%-reproducible timeout bug in `execute_command`. The polling loop detected completion by reading `finished_at`, a field the Alpacon Command API never returns, so a command that finished instantly was never recognized as complete and the loop polled until the client timed out. The fix switches completion detection to `handled_at`, adds a structured `pending_approval` short-circuit for `awaiting_approval`, folds `denied`/`rejected` into the existing `stuck`/`error` fast-fail branch, registers `COMMAND_AWAITING_APPROVAL` in `_NEXT_ACTION_BY_CATEGORY`, and corrects a `list_commands` description that pointed at a nonexistent tool.

I verified the central claim directly against the alpacon-server source in the sibling checkout:

- **`handled_at` is the correct completion signal.** `Command.compute_status` (`events/models.py:659-660`) returns `'success'`/`'failed'` exactly when `handled_at is not None` (after the `denied`/`rejected` guards). So `handled_at is not None` ⟺ a terminal, result-bearing state. The old `finished_at` field does not exist anywhere in `events/` (confirmed by grep) — the check never matched since day one, exactly as the PR body states.
- **No premature-completion race.** `Command.fin()` (`events/models.py:924-934`) writes `success`, `exit_code`, `result`, `masked_result`, `elapsed_time`, and `handled_at` in a single `.update()` inside `transaction.atomic()`, guarded by `handled_at__isnull=True`. When the MCP reads a row with `handled_at` set, the result payload is already populated — there is no window where `handled_at` is visible before the output.
- **The API exposes the fields the fix relies on.** `CommandSerializer` (`events/api/serializers.py:102-176`) serializes `handled_at`, `success`, `exit_code`, and a `status` computed via `compute_status(now)` at read time. The retrieve endpoint `/api/events/commands/{id}/` returns exactly what the loop reads.
- **The terminal-state routing matches the server.** `denied` (ACL), `rejected` (HITL reject / reaper-flipped `awaiting_approval`, `events/tasks.py:167-204`), and `stuck` are all states where `handled_at` stays null and the command never returns a result — fast-failing is correct and matches pre-PR behavior. `awaiting_approval` in `compute_status` requires `delivered_at is None`, so short-circuiting it to `pending_approval` before delivery is right.

The completion-detection fix is correct and well-supported by the server contract. My findings are about the agent-facing guidance text, not the bug fix itself.

### Critical Issues (0)
None. The core behavior change is semantically correct and verified against the server source.

### Warnings (2)

1. **[tools/command_tools.py:346]** The `awaiting_approval` message advises "check the result via list_commands **or re-run**." For this state the original command is *pending delivery* and will execute once a human approves — HITL approval sets `verification_status='approved'` and calls `command.execute()` (`events/services.py:223-238`), after which the same command reaches `handled_at`. "Re-run" therefore submits a *duplicate* command that needs its own separate approval, and (if a human approves both) can double-execute an already-risk-flagged HITL command. The correct recovery is to wait and retrieve via `list_commands`. Unlike the sudo-denial categories — where nothing was delivered and re-running genuinely is the recovery — re-run is counterproductive here. Recommend dropping "or re-run" from this message. Non-blocking.

2. **[utils/common.py:95]** Same issue in the registered `COMMAND_AWAITING_APPROVAL` next_action: "...check the result via list_commands or re-run; do not repeatedly resubmit." The "or re-run" clause is in tension with the "do not repeatedly resubmit" clause in the same sentence, and for `awaiting_approval` the command runs on approval. Recommend "Wait for approval, then retrieve the result via list_commands; do not resubmit." Non-blocking.

### Suggestions (3)

1. **[tools/command_tools.py:370]** (pre-existing, not changed by this PR) Now that completion detection actually works, the in-progress states the server can return — `verifying` (AI-verdict pending), `delivered`, `queued`, `scheduled` — poll against the *original* deadline, because only `status in ('running', 'acked')` resets it. `'acked'` in that tuple is also dead: `compute_status` never returns `'acked'` (it returns `'running'` once `acked_at` is set). A slow AI verification could time `execute_command` out while the command later completes. This is acceptable given the documented submit-then-retrieve recovery, but consider treating `verifying` like `running` for the deadline reset, or documenting that pre-execution states count against the original timeout.

2. **[tools/command_tools.py:358]** The terminal tuple includes `'error'`, which `compute_status` marks unreachable (`# type: ignore[unreachable]` at `events/models.py:680`) — the model always has a `scheduled_at` default. Harmless defensive coding, but it pins a status the server can't emit; worth a brief comment or removal for clarity.

3. **[tests/test_command_tools.py]** Every completion test returns `handled_at` on the *first* poll. There's no test of the multi-poll transition (poll 1 = `running`/`verifying` with `handled_at: None`, poll 2 = `handled_at` set), which is the exact path the bug broke. `test_timeout` covers "keeps polling then times out," but a transition-to-completion test would directly pin the fixed behavior. Optional.

### Highlights
- **Security-correct pending path.** The `awaiting_approval` branch deliberately does *not* forward `data=result`, so the raw poll payload (which can carry AI risk internals like `risk_score`/`tier2_reasoning`) is never disclosed — only the category and next_action, consistent with ADR 0015 and the sudo-denial path.
- **Root cause verified, not guessed.** The PR body's live reproduction (submit-only path returns, polling path times out) correctly isolates the wait path, and the `finished_at`-never-existed claim holds against the source.
- **Phantom-tool doc fix is real.** There is no `get_command_result` MCP tool; the old `list_commands` "Related" pointer would have sent an agent to a nonexistent tool. The new text points at the genuine recovery path.
- **Fixtures moved to the real API shape** (`handled_at` + `success`/`failed`, dropping `finished_at`/`completed`), so the tests now pin the actual contract instead of the fiction that hid the bug.

**Verdict**: APPROVED

jisung-02:
Thanks for the thorough review, @junho226. Addressed all five points in 4aab421:

**Warning 1 & 2 — "or re-run" on awaiting_approval.** Agreed this is counterproductive: a command in `awaiting_approval` runs automatically once a human approves, so re-running submits a duplicate that needs its own approval and risks double-executing a HITL-flagged command. Removed "or re-run" from both the `execute_command` message (`tools/command_tools.py`) and the `COMMAND_AWAITING_APPROVAL` next_action (`utils/common.py`); both now instruct the agent to wait and retrieve via `list_commands`, and explicitly warn against re-running.

**Suggestion 1 — in-progress states poll against the original deadline.** The deadline reset now covers all in-progress states (`running`, `verifying`, `delivered`, `queued`, `scheduled`), so a slow AI verification or delayed delivery no longer times out a command that is still advancing. Dropped the dead `acked` (compute_status returns `running` once acked_at is set).

**Suggestion 2 — unreachable `error`.** Removed `error` from the terminal fast-fail tuple and added a comment noting compute_status never emits it given the `scheduled_at` default.

**Suggestion 3 — multi-poll transition test.** Added `test_completes_after_polling`: poll 1 = `running`/handled_at=None, poll 2 = `verifying`/handled_at=None, poll 3 = `handled_at` set. Asserts completion is detected on the transition (call_count == 3), directly pinning the path the original bug broke.

junho226 (approved):

PR #138, opened 2026-07-24
feat: add --toolsets for selective tool registration in local mode (#34)

### LINKED ISSUE #34: feat: toolset grouping for tool count management
## Summary

Add `--toolsets` CLI argument for local (stdio/SSE) mode to selectively enable tool groups. Remote (streamable-http) mode registers all tools and relies on Claude's automatic Tool Search optimization.

## Background

- Anthropic recommends max 30-50 tools for optimal tool selection accuracy
- Each tool definition costs ~1.2K tokens per API request
- As new tools are added (Phase 1-3), total count will exceed 100
- GitHub MCP Server uses similar toolset grouping pattern

## Implementation

- Add `TOOLSET_REGISTRY` mapping toolset names to tool modules in `server.py`
- Add `DEFAULT_TOOLSETS` (servers, commands, websh, metrics, events)
- Add `_resolve_toolsets()` function to parse CLI argument
- In `run()`: remote mode → register all tools, local mode → selective import based on toolsets
- Add `--toolsets` arg to `main.py`, `main_sse.py`
- `workspace_tools` always registered regardless of toolset selection

## Usage

```json
{
  "mcpServers": {
    "alpacon": {
      "command": "uvx",
      "args": ["alpacon-mcp", "--toolsets", "servers,websh,metrics,iam"]
    }
  }
}
```

Default: `servers,commands,websh,metrics,events`
Use `--toolsets all` to enable everything.

jisung-02:
## Background

Local stdio/SSE mode registered all 19 tool modules unconditionally in `run()`. Clients cap the number of exposed tools, and more registered tools raise the per-request token cost, but there was no way to register only the toolsets a workflow actually uses.

## What changed

Add a `--toolsets` CLI argument and an `ALPACON_MCP_TOOLSETS` environment variable to select which toolsets local mode registers. Precedence is CLI argument > env var > default `all`, so the default keeps existing behavior unchanged.

Tool registration is an import-time side effect of the `@mcp_tool_handler` decorator, so selection has to happen at the import boundary. The hardcoded import block in `run()` (19 tool modules plus a `tools.resources` import) was replaced with an `importlib.import_module` loop driven by `TOOLSET_REGISTRY` (15 toolset names mapped 1:1 to modules). `resolve_toolsets` resolves and validates the names, matching them case-insensitively and rejecting unknown names before the `all` short-circuit so a typo alongside `all` still fails. An unknown name raises `ToolsetError` (a `ValueError` subclass), which both entry points catch to log a one-line `Invalid --toolsets` message and exit 2 — scoped to that type so an unrelated `ValueError` raised while importing a tool module keeps the full-traceback path. `_modules_to_load` decides the final module set based on remote vs local.

`workspace`, `health`, `work-sessions`, and `prompts` are always registered. `work_session_tools` in particular must stay, because gate-denial responses instruct the agent to call `work_session_*`. These four names are accepted in `--toolsets` but have no effect on selection; any other unrecognized name fails at startup. If a selection resolves to only always-on names (e.g. `--toolsets health`), it registers just the always-on tools and logs a warning so the all-no-op selection is not silent.

`tools/resources.py` was refactored into a registry that references tools as `'module.func'` strings, so importing it no longer imports (and registers) tool modules, which would defeat selection. `register_resources(enabled_modules)` registers only resources whose backing module is enabled, and `run()` now calls it explicitly since resource registration is no longer an import side effect. A few resources cross module boundaries — `alpacon://servers/.../overview` is backed by `system_info_tools` — so the README notes that `--toolsets servers` alone drops it and needs `system-info` added.

The broken `alpacon-mcp-sse` console script was fixed along the way: it pointed at `main_sse:main`, which did not exist, so a `main()` was added to `main_sse.py`.

## Scope

Remote mode ignores `--toolsets` and always registers every tool; it relies on the client's tool-search optimization, and this avoids accidentally narrowing tools in an authenticated remote deployment. "Remote mode" is gated on `ALPACON_MCP_AUTH_ENABLED=true` (JWT auth), not on the transport name, matching the existing `_is_remote_mode` definition. When a toolsets value is detected in remote mode, it is logged as ignored.

Narrow selections cut real capability: the workflow prompts still reference tools such as `execute_command` and `webftp_upload_file`, so an agent following them will hit "tool not found" if you exclude `commands` or `webftp`. Select the toolsets your workflows actually use.

## Tests

New `tests/test_toolsets.py` covers `resolve_toolsets` (precedence, whitespace, case-insensitive matching, unknown-name fail-fast raising `ToolsetError`, always-on names ignored, always-on-only selection warns and selects nothing), `_modules_to_load` (local subset plus always-on, remote full set, local default full set), registry-to-disk consistency, an AST guard that `resources.py` imports no tool module, `register_resources` filtering, both entry points forwarding `--toolsets`, and `run()` wiring.

Full suite: 949 passed. Two failures (`test_basic.py::test_import_main`, `test_simple_integration.py::TestMCPServerConfiguration::test_main_entry_point`) are pre-existing and unrelated to this branch: the checkout directory is named `main`, which shadows the top-level `main.py` on `import main` (they reproduce on the base commit). ruff lint and format pass.

Closes #34

### DIFF
--- a/README.md
+++ b/README.md
@@ -189,7 +189,44 @@ uvx alpacon-mcp setup --token-file ~/my.json   # Use custom file location
 uvx alpacon-mcp test                           # Test your connection
 uvx alpacon-mcp list                           # Show configured workspaces
 uvx alpacon-mcp add                            # Add another workspace (shows path)
+uvx alpacon-mcp --toolsets servers,commands,webftp   # Register only these toolsets
 ```
+
+---
+
+### Toolsets (selective tool registration)
+
+Local (stdio/SSE) mode can register a subset of toolsets to stay within
+client tool limits and reduce per-request token cost. Remote
+(streamable-http) mode always registers all tools and relies on the
+client's tool search optimization.
+
+```json
+{
+  "mcpServers": {
+    "alpacon": {
+      "command": "uvx",
+      "args": ["alpacon-mcp", "--toolsets", "servers,commands,webftp,metrics"]
+    }
+  }
+}
+```
+
+The `ALPACON_MCP_TOOLSETS` environment variable is also supported; the CLI
+argument wins when both are set. Default: `all` (register everything).
+
+Available toolsets (1:1 with tool modules): `servers`, `commands`, `webftp`,
+`metrics`, `alerts`, `events`, `system-info`, `iam`, `security`, `audit`,
+`approvals`, `webhooks`, `packages`, `certs`, `tokens`. Workspace, health,
+and Work Session tools are always registered regardless of selection, as are
+the MCP prompts; their names (`workspace`, `health`, `work-sessions`,
+`prompts`) are accepted in `--toolsets` but have no effect. Any other
+unrecognized name fails at startup.
+
+Narrow selections cut real capability: the workflow prompts still reference
+tools such as `execute_command` and `webftp_upload_file`, so an agent
+following them will hit "tool not found" if you excluded `commands` or
+`webftp`. Select the toolsets your workflows actually use.
 
 ---
 
--- a/README.md
+++ b/README.md
@@ -197,9 +197,10 @@ uvx alpacon-mcp --toolsets servers,commands,webftp   # Register only these tools
 ### Toolsets (selective tool registration)
 
 Local (stdio/SSE) mode can register a subset of toolsets to stay within
-client tool limits and reduce per-request token cost. Remote
-(streamable-http) mode always registers all tools and relies on the
-client's tool search optimization.
+client tool limits and reduce per-request token cost. Remote mode
+(streamable-http with JWT auth, i.e. `ALPACON_MCP_AUTH_ENABLED=true`)
+always registers all tools and relies on the client's tool search
+optimization.
 
 ```json
 {
--- a/README.md
+++ b/README.md
@@ -227,7 +227,10 @@ unrecognized name fails at startup.
 Narrow selections cut real capability: the workflow prompts still reference
 tools such as `execute_command` and `webftp_upload_file`, so an agent
 following them will hit "tool not found" if you excluded `commands` or
-`webftp`. Select the toolsets your workflows actually use.
+`webftp`. Select the toolsets your workflows actually use. A few resources
+also cross module boundaries: `alpacon://servers/.../overview` is backed by
+`system_info_tools`, so `--toolsets servers` alone silently drops it—add
+`system-info` to keep it.
 
 ---
 
--- a/main.py
+++ b/main.py
@@ -114,6 +114,10 @@ Examples:
 
     try:
         run('stdio', config_file=args.config_file, toolsets=args.toolsets)
+    except ValueError as e:
+        # A toolsets typo is user error, not a crash: one clean line, no traceback.
+        logger.error(f'Invalid --toolsets: {e}')
+        raise SystemExit(2)
     except Exception as e:
         logger.error(f'Failed to start MCP server: {e}', exc_info=True)
         raise
--- a/main.py
+++ b/main.py
@@ -2,7 +2,7 @@
 import argparse
 from pathlib import Path
 
-from server import run
+from server import TOOLSETS_HELP, run
 from utils.logger import get_logger
 
 logger = get_logger('main')
@@ -35,6 +35,7 @@ Examples:
   uvx alpacon-mcp test               # Test connection
   uvx alpacon-mcp list               # Show configured workspaces
   uvx alpacon-mcp add                # Add another workspace
+  uvx alpacon-mcp --toolsets servers # Register only the listed toolsets
         """,
     )
     parser.add_argument(
@@ -57,6 +58,11 @@ Examples:
         '--token-file',
         type=str,
         help='Custom path to token.json file (overrides --local and default locations)',
+    )
+    parser.add_argument(
+        '--toolsets',
+        type=str,
+        help=TOOLSETS_HELP,
     )
 
     args = parser.parse_args()
@@ -107,7 +113,7 @@ Examples:
     logger.info(f'Configuration: config_file={args.config_file}')
 
     try:
-        run('stdio', config_file=args.config_file)
+        run('stdio', config_file=args.config_file, toolsets=args.toolsets)
     except Exception as e:
         logger.error(f'Failed to start MCP server: {e}', exc_info=True)
         raise
--- a/main.py
+++ b/main.py
@@ -2,7 +2,7 @@
 import argparse
 from pathlib import Path
 
-from server import TOOLSETS_HELP, run
+from server import TOOLSETS_HELP, ToolsetError, run
 from utils.logger import get_logger
 
 logger = get_logger('main')
@@ -114,8 +114,10 @@ Examples:
 
     try:
         run('stdio', config_file=args.config_file, toolsets=args.toolsets)
-    except ValueError as e:
+    except ToolsetError as e:
         # A toolsets typo is user error, not a crash: one clean line, no traceback.
+        # Scoped to ToolsetError so an unrelated ValueError during tool import
+        # (e.g. a bad numeric env var) still gets the full-traceback path below.
         logger.error(f'Invalid --toolsets: {e}')
         raise SystemExit(2)
     except Exception as e:
--- a/main_sse.py
+++ b/main_sse.py
@@ -1,10 +1,10 @@
 # main_sse.py
 import argparse
-import logging
 
-from server import TOOLSETS_HELP, run
+from server import TOOLSETS_HELP, ToolsetError, run
+from utils.logger import get_logger
 
-logger = logging.getLogger(__name__)
+logger = get_logger('main_sse')
 
 
 def main():
@@ -24,8 +24,10 @@ def main():
     args = parser.parse_args()
     try:
         run('sse', config_file=args.config_file, toolsets=args.toolsets)
-    except ValueError as e:
+    except ToolsetError as e:
         # A toolsets typo is user error, not a crash: one clean line, no traceback.
+        # Scoped to ToolsetError so an unrelated ValueError during tool import is
+        # not mislabeled as an --toolsets problem.
         logger.error(f'Invalid --toolsets: {e}')
         raise SystemExit(2)
 
--- a/main_sse.py
+++ b/main_sse.py
@@ -1,16 +1,26 @@
 # main_sse.py
 import argparse
 
-from server import run
+from server import TOOLSETS_HELP, run
 
-# Entry point to run the server with SSE
-if __name__ == '__main__':
+
+def main():
+    """Entry point for the alpacon-mcp-sse console script."""
     parser = argparse.ArgumentParser(description='Alpacon MCP Server (SSE mode)')
     parser.add_argument(
         '--config-file',
         type=str,
         help='Path to token configuration file (overrides default config discovery)',
     )
+    parser.add_argument(
+        '--toolsets',
+        type=str,
+        help=TOOLSETS_HELP,
+    )
 
     args = parser.parse_args()
-    run('sse', config_file=args.config_file)
+    run('sse', config_file=args.config_file, toolsets=args.toolsets)
+
+
+if __name__ == '__main__':
+    main()
--- a/main_sse.py
+++ b/main_sse.py
@@ -1,7 +1,10 @@
 # main_sse.py
 import argparse
+import logging
 
 from server import TOOLSETS_HELP, run
+
+logger = logging.getLogger(__name__)
 
 
 def main():
@@ -19,7 +22,12 @@ def main():
     )
 
     args = parser.parse_args()
-    run('sse', config_file=args.config_file, toolsets=args.toolsets)
+    try:
+        run('sse', config_file=args.config_file, toolsets=args.toolsets)
+    except ValueError as e:
+        # A toolsets typo is user error, not a crash: one clean line, no traceback.
+        logger.error(f'Invalid --toolsets: {e}')
+        raise SystemExit(2)
 
 
 if __name__ == '__main__':
--- a/server.py
+++ b/server.py
@@ -1,3 +1,4 @@
+import importlib
 import os
 import signal
 import sys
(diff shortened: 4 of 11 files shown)


tests/test_resources.py
```diff
@@ -1,8 +1,19 @@
 """Tests for alpacon:// MCP resources."""
 
+import inspect
+import re
+
 import pytest
 
-from server import mcp
+import tools.resources as res
```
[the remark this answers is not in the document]
jisung-02: Fixed in d419dc5. Removed the `from tools.resources import register_resources` import and call `res.register_resources(...)` through the existing module alias, so the file now uses a single import style for `tools.resources`.

tests/test_toolsets.py
```diff
@@ -0,0 +1,217 @@
...
+
+import ast
+import importlib
+import importlib.util
+import sys
+from pathlib import Path
+
+import pytest
+
+import server
```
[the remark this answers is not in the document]
jisung-02: Fixed in d419dc5. Removed the `from server import (...)` block and reference every name through `server.<name>`, keeping the required `import server` (needed for `server.__file__`, `server.mcp`, `server.run`, and monkeypatching `server.importlib`). Also scanned the rest of the branch's changed files: no other module mixes both import styles.

server.py
```diff
@@ -157,6 +158,85 @@ def _create_mcp_server() -> FastMCP:
...
+        for n in names
+        if n != TOOLSETS_ALL
+        and n not in TOOLSET_REGISTRY
+        and n not in ALWAYS_ON_TOOLSET_NAMES
+    ]
+    if unknown:
+        valid = ', '.join([*sorted(TOOLSET_REGISTRY), TOOLSETS_ALL])
+        raise ValueError(
+            f'Unknown toolset(s): {", ".join(unknown)}. Valid toolsets: {valid}'
+        )
```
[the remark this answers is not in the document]
jisung-02: Good catch, fixed in d419dc5. The startup error now lists the always-on names (`workspace`, `health`, `work-sessions`, `prompts`) alongside the selectable toolsets and `all`, so a typo of an accepted name surfaces it as valid. Added a test asserting an always-on name appears in the message.

main.py
```diff
@@ -107,7 +113,7 @@ def main():
     logger.info(f'Configuration: config_file={args.config_file}')
 
     try:
-        run('stdio', config_file=args.config_file)
+        run('stdio', config_file=args.config_file, toolsets=args.toolsets)
```
junho226, line 116: Suggestion (optional, UX): a bad `--toolsets` value — or a bad `ALPACON_MCP_TOOLSETS` env value — raises `ValueError` from `resolve_toolsets`, which lands in the `except Exception` just below and is logged with `exc_info=True`. So a simple toolset typo surfaces as a full traceback in the client's server log.

The `ValueError` message is already clean and actionable, and raising in `resolve_toolsets` is the right place (the env-var path means argparse validation can't cover it) — this is only about presentation. Consider catching it separately for a one-line startup error:

```suggestion
        run('stdio', config_file=args.config_file, toolsets=args.toolsets)
    except ValueError as e:
        logger.error(f'Invalid --toolsets: {e}')
        raise SystemExit(2)
```

README.md
```diff
@@ -189,10 +189,48 @@ uvx alpacon-mcp setup --token-file ~/my.json   # Use custom file location
...
+      "args": ["alpacon-mcp", "--toolsets", "servers,commands,webftp,metrics"]
+    }
+  }
+}
+```
+
+The `ALPACON_MCP_TOOLSETS` environment variable is also supported; the CLI
+argument wins when both are set. Default: `all` (register everything).
+
+Available toolsets (1:1 with tool modules): `servers`, `commands`, `webftp`,
```
junho226, line 219: Suggestion (optional, docs): resources are backed 1:1 by tool *modules*, but a few live under a different namespace's URI — e.g. `server_overview` (`alpacon://servers/.../overview`) is backed by `system_info_tools`, so `--toolsets servers` alone silently drops that one resource. The "1:1 with tool modules" line plus the capability warning below mostly cover this, but a one-line note that some `servers/*` resources also need `system-info` would preempt confusion.

junho226 (dismissed):
## Review: `--toolsets` selective tool registration (#34)

Verified locally in a PR worktree: `tests/test_toolsets.py` + `test_resources.py` + `test_health.py` → **42 passed (0.50s)**.

### Summary
Adds `--toolsets` / `ALPACON_MCP_TOOLSETS` so local (stdio/SSE) mode registers only selected tool modules, cutting exposed tool count and per-request token cost. The hardcoded 19-module import block in `run()` becomes an `importlib` loop driven by `TOOLSET_REGISTRY` (15) + `ALWAYS_ON` (4); `tools/resources.py` is refactored to reference tools as `module.func` strings so importing it no longer eagerly registers tool modules. Default stays `all` (non-breaking); remote mode ignores the setting. Also fixes the broken `alpacon-mcp-sse` console script.

### Critical / Warnings
None.

### Suggestions (optional, non-blocking)
See the two inline comments (startup-error presentation, and a cross-module resource note in the README).

### Highlights
- **Injection-safe dynamic import**: `--toolsets` input is validated against the registry and mapped to *hardcoded* module names before `importlib.import_module` — a user-controlled string is never imported.
- **Validate-before-short-circuit**: `all,typo` still fails fast, with a dedicated test.
- **Regression-proofed**: the AST guard (`resources.py` imports no tool module) locks the core design invariant, and the registry-to-disk consistency test guarantees the refactor drops no tool module (19 old imports = 15 registry + 4 always-on).
- Bot feedback (Copilot / github-code-quality, 3 items) already resolved in d419dc5.
- Default `all` non-breaking; the `run()` integration seam is covered end-to-end.

**Verdict: APPROVE** — clean, well-structured, thoroughly tested.

main.py
jisung-02: Fixed in 14b8cd3. `main.py` now catches `ValueError` from `resolve_toolsets` separately and logs a one-line `Invalid --toolsets: ...` then exits with `SystemExit(2)`, so a toolset typo no longer surfaces as a full traceback. Applied the same guard to `main_sse.py`, which previously had no try/except and let the `ValueError` propagate raw. The raise stays in `resolve_toolsets` (the env-var path is outside argparse), caught at each entry point.

README.md
jisung-02: Fixed in 14b8cd3. Added a one-line note to the README toolsets section: `alpacon://servers/.../overview` is backed by `system_info_tools`, so `--toolsets servers` alone silently drops it and you need to add `system-info` to keep it.

main.py
```diff
@@ -107,7 +113,11 @@ def main():
...
+    except ValueError as e:
```
junho226: Warning: this catch wraps the entire run() call, which imports the tool modules, so any ValueError raised during module import (not just a toolsets typo) is reported as "Invalid --toolsets". Concrete case: tools/webftp_tools.py:45 does int(os.getenv('ALPACON_MCP_WEBFTP_DOWNLOAD_TIMEOUT', '60')) at import time, so ALPACON_MCP_WEBFTP_DOWNLOAD_TIMEOUT=abc makes this print "Invalid --toolsets: invalid literal for int()..." even with no --toolsets set. A malformed token.json (json.JSONDecodeError, a ValueError subclass) is mislabeled the same way. Validate the toolsets value before run() (call resolve_toolsets/_modules_to_load in the entry point and catch there), or raise a dedicated ToolsetError(ValueError) from resolve_toolsets and catch that specific type.

main_sse.py
```diff
@@ -1,16 +1,34 @@
...
+        '--toolsets',
+        type=str,
+        help=TOOLSETS_HELP,
+    )
 
     args = parser.parse_args()
-    run('sse', config_file=args.config_file)
+    try:
+        run('sse', config_file=args.config_file, toolsets=args.toolsets)
+    except ValueError as e:
```
junho226: Warning: same issue as main.py:117 — this except ValueError wraps all of run(), so any ValueError from tool-module import (e.g. int(os.getenv('ALPACON_MCP_WEBFTP_DOWNLOAD_TIMEOUT')) in webftp_tools, or json.JSONDecodeError from a bad token.json) is mislabeled as "Invalid --toolsets". Fix both entry points together: scope the catch to the actual resolution step or catch a dedicated ToolsetError.

main_sse.py
```diff
@@ -1,16 +1,34 @@
 # main_sse.py
 import argparse
+import logging
 
-from server import run
+from server import TOOLSETS_HELP, run
 
-# Entry point to run the server with SSE
-if __name__ == '__main__':
+logger = logging.getLogger(__name__)
```
junho226: Suggestion: main.py uses the project logger (from utils.logger import get_logger), but this uses the stdlib logging.getLogger(__name__). The "Invalid --toolsets" error then bypasses the project's formatter/handlers and only reaches stderr via the last-resort handler. Use get_logger('main_sse') for consistency and correct log routing.

server.py
```diff
@@ -157,6 +158,96 @@ def _create_mcp_server() -> FastMCP:
...
+    'prompts': 'prompts',
+}
+ALWAYS_ON_MODULES: frozenset[str] = frozenset(ALWAYS_ON.values())
+ALWAYS_ON_TOOLSET_NAMES: frozenset[str] = frozenset(ALWAYS_ON)
+
+
+def resolve_toolsets(toolsets: str | None) -> set[str]:
+    """CLI arg > ALPACON_MCP_TOOLSETS env var > 'all'; unknown name -> ValueError."""
+    raw = toolsets if toolsets is not None else os.getenv(TOOLSETS_ENV_VAR)
+    names = [n.strip() for n in raw.split(',') if n.strip()] if raw else []
```
junho226: Suggestion: toolset name matching is case-sensitive, so --toolsets Servers or SERVERS raises "Unknown toolset(s)" even though every registry key is lowercase. Since the accepted vocabulary is entirely lowercase, normalizing here with n.strip().lower() would be friendlier with no ambiguity cost.

junho226 (commented):
### Summary

PR #138 adds `--toolsets` (and the `ALPACON_MCP_TOOLSETS` env var) so local stdio/SSE mode can register a subset of tool modules, to stay within client tool limits and cut per-request token cost. The design is coherent and the implementation is the right shape:

- `server.py` gains `TOOLSET_REGISTRY` (15 toolset names -> tool modules 1:1), an `ALWAYS_ON` set (`workspace`, `health`, `work-sessions`, `prompts`), `resolve_toolsets()` (CLI arg > env var > `all`, unknown name -> `ValueError`), and `_modules_to_load()` (remote mode forces all + logs that it ignores the setting).
- `run()` replaces the static import block with a dynamic `importlib.import_module` loop over the resolved set, then calls the new `register_resources(modules)`.
- `tools/resources.py` is the crux: it drops eager `from tools.x import ...` imports and stores tool refs as lazy `'module.func'` strings resolved via `importlib`, so importing the resource registry no longer registers every tool. Registration moves from a module-level loop into `register_resources(enabled_modules)`, which filters resources by their backing module.

The registration gating is genuine (a non-selected module is never imported, so its `@mcp.tool` decorators never run), backward compatibility is preserved (default with no flag/env resolves to every module, identical to before), and there is no import-injection surface (user names are validated against a fixed allow-list; `import_module` only ever receives hardcoded registry values). Test coverage is excellent and the docs even anticipate the cross-module resource pitfall.

One robustness issue is worth addressing before merge (misleading error labeling), plus a few minor suggestions. No blockers.

### Critical Issues (0)

None.

### Warnings (1)

1. **[main.py:117] / [main_sse.py:27]** The `except ValueError` wraps the *entire* `run()` call, which imports the tool modules (each `import_module` executes module-level code). Any `ValueError` raised during that import is mislabeled as `Invalid --toolsets` and exits 2. Concrete case: `tools/webftp_tools.py:45` does `int(os.getenv('ALPACON_MCP_WEBFTP_DOWNLOAD_TIMEOUT', '60'))` at import time, so `ALPACON_MCP_WEBFTP_DOWNLOAD_TIMEOUT=abc` makes `uvx alpacon-mcp` (with or without `--toolsets`) print `Invalid --toolsets: invalid literal for int()...`. A malformed `token.json` raises `json.JSONDecodeError` (a `ValueError` subclass) and is reported the same way. Scope the catch to the actual resolution step (validate before `run()`), or raise a dedicated `ToolsetError(ValueError)` and catch that.

### Suggestions (3)

1. **[server.py:204]** Toolset name matching is case-sensitive, so `--toolsets Servers`/`SERVERS` fails as "Unknown toolset(s)" though every registry key is lowercase. Normalizing with `n.strip().lower()` would be friendlier with no ambiguity cost.
2. **[main_sse.py:7]** This entry point uses stdlib `logging.getLogger(__name__)`, while `main.py` uses the project's `get_logger`. The error then bypasses the project's formatter/handlers. Use `get_logger('main_sse')` for consistency.
3. **[tests/test_http_client.py]** The changes here are pure `ruff format` reflows unrelated to `--toolsets`; harmless, but bundling formatting churn into a feature PR adds review noise.

### Highlights

- **Genuine gating, verified backward-compat.** Default resolves to `ALL_TOOL_MODULES | ALWAYS_ON_MODULES`, exactly the modules the old static block imported; `test_registry_covers_every_tool_module` fails CI if a new tool module forgets registry wiring.
- **The lazy `'module.func'` ref indirection in `resources.py` is the correct core fix** — it decouples "list the resource registry" from "import (and register) the backing tool," backed by an AST test that fails if anyone reintroduces an eager tool import.
- **Validation ordering is deliberate**: unknown names are checked before the `all` short-circuit, so `all,typo` still fails.
- **Cross-module resource ordering is preserved** — `register_resources` sorts the filtered set by placeholder count in one pass, keeping the literal-before-wildcard invariant.
- **Docs are honest about sharp edges**: the README documents that narrow selections break prompt-referenced tools and that `alpacon://servers/.../overview` is backed by `system_info_tools`.
- **Strong tests**: resolve edge cases (empty, `all`, whitespace, CLI-vs-env precedence, unknown fail-fast, always-on aliases), a `run()` integration test catching a parts-mismatch, and a ref-resolves-to-real-function check.

---

**Verdict**: COMMENTED

main.py
jisung-02: Good catch, fixed in 28f85b4. Introduced `ToolsetError(ValueError)`, raised only by `resolve_toolsets`, and changed this handler to `except ToolsetError`. An unrelated `ValueError` raised while importing a tool module (e.g. `int(os.getenv("ALPACON_MCP_WEBFTP_DOWNLOAD_TIMEOUT"))` in `webftp_tools` at import time) is now no longer mislabeled as `Invalid --toolsets` and keeps the full-traceback `except Exception` path below. I scoped it with a typed exception rather than narrowing the try block, so it stays correct regardless of where inside `run()` the toolset validation happens.

main_sse.py
jisung-02: Fixed in 28f85b4. Same `ToolsetError` scoping applied here, so a `ValueError` from tool-module import or a bad `token.json` is no longer reported as an `--toolsets` problem. main_sse.py has no generic `except Exception` (it only guards toolset input), so unrelated errors now propagate raw instead of being swallowed under the wrong label.

main_sse.py
jisung-02: Fixed in 28f85b4. Switched to `from utils.logger import get_logger` and `get_logger("main_sse")`, matching main.py, so the startup error goes through the project formatter/handlers instead of the stdlib last-resort stderr handler.

server.py
jisung-02: Fixed in 28f85b4. `resolve_toolsets` now normalizes each name with `n.strip().lower()`, so `--toolsets Servers` or `SERVERS` resolves like `servers`. Added a `test_names_are_case_insensitive` case. Since the accepted vocabulary is entirely lowercase there is no ambiguity.

junho226 (approved):
### Re-review (delta only)

Scoped to the single commit since my 2026-07-24 review (`28f85b4`, 4 files, +27 −10). It addresses every point I raised.

**Resolved**

- **Warning 1 (mislabeled startup error)** — the new `ToolsetError(ValueError)` plus the narrowed catch in `main.py`/`main_sse.py` is exactly the right shape: subclassing `ValueError` keeps the existing contract while letting the entry points catch only their own error. Verified the original repro: with `ALPACON_MCP_WEBFTP_DOWNLOAD_TIMEOUT=abc`, importing `tools.webftp_tools` now raises a plain `ValueError`, which no longer matches `except ToolsetError` and falls through to the full-traceback path. A malformed `token.json` (`json.JSONDecodeError`) is handled the same way.
- **Suggestion 1 (case sensitivity)** — `.strip().lower()` normalization, with `test_names_are_case_insensitive` covering it. Safe because every `TOOLSET_REGISTRY`/`ALWAYS_ON` key is lowercase; `ALL` resolves to all 15 modules too.
- **Suggestion 2 (logger consistency)** — `main_sse.py` now uses `get_logger('main_sse')`.

**Verification** — `tests/test_toolsets.py` + `test_resources.py` + `test_health.py` → **43 passed** (was 42). `ruff check` and `ruff format --check` both clean.

**Nit (non-blocking)** — the error echoes the normalized name, not the user's input: `--toolsets Nope` reports `Unknown toolset(s): nope`. Keeping the raw token in the message (validating against the lowercased form) would make a typo easier to spot. Trivial.

`tests/test_http_client.py` still carries the unrelated formatting reflow noted last time — harmless, just review noise.

**Verdict: APPROVE**

PR #139, opened 2026-07-28
refactor(utils): consolidate ALPACON_MCP_AUTH_ENABLED check into one helper

jisung-02:
## Background

The remote-mode check `os.getenv('ALPACON_MCP_AUTH_ENABLED', '').lower() == 'true'` was copy-pasted at five sites: `utils/health.py:17`, `utils/http_client.py:177`, `utils/decorators.py:33`, and twice in `server.py` (lines 77 and 259). Two of those five were wrapped in separate, identically-bodied private helpers that happened to share a name — `_is_remote_mode()` existed once in `utils/health.py` and again in `server.py`. On top of that, `tools/webftp_tools.py:20` reached across a module boundary to import the module-private `utils.decorators._is_auth_enabled` and called it at four sites.

One env var deciding local (stdio/SSE, token.json) versus remote (streamable-http, Auth0 JWT) mode should be read through one function.

## What changed

`is_auth_enabled()` now lives in `utils/common.py` and is the only place that reads `ALPACON_MCP_AUTH_ENABLED`. All five original checks call it, and the duplicated private helpers (`utils.decorators._is_auth_enabled`, `utils.health._is_remote_mode`, `server._is_remote_mode`) are gone.

| File | Before | After |
|---|---|---|
| `utils/common.py` | — | defines `is_auth_enabled()` |
| `utils/decorators.py` | `_is_auth_enabled()` helper | imports `is_auth_enabled` |
| `utils/health.py` | `_is_remote_mode()` helper | imports `is_auth_enabled` |
| `server.py` | inline check + `_is_remote_mode()` helper | imports `is_auth_enabled` |
| `utils/http_client.py` | inline check | imports `is_auth_enabled` |
| `tools/webftp_tools.py` | imports private `_is_auth_enabled` from `utils.decorators` | imports `is_auth_enabled` from `utils.common` |
| `tools/workspace_tools.py` | function-local import of private `_is_auth_enabled` | module-level import of `is_auth_enabled` |

`utils/common.py` was chosen as the home because it already hosts the shared helpers and sits below every consumer in the import graph — it only imports `utils.logger` and `utils.token_manager`. Hosting the helper in `utils/decorators.py` was not an option: `utils/decorators.py:26` imports `utils.http_client`, so `http_client` importing back from `decorators` would be a cycle.

Two follow-on edits: `utils/http_client.py` drops its `os` import, which the replaced line was the last user of, and the seven `patch(...)` targets in `tests/test_webftp_tools.py` move from `tools.webftp_tools._is_auth_enabled` to `tools.webftp_tools.is_auth_enabled`.

One small tidy-up rides along in a separate commit: `utils/common.py` declarations are reordered — module constants collected into one block after the imports, functions following in dependency order (define before use), private helper last. Pure code motion; with blank lines dropped the line multiset is identical before and after, and no signature, body, or import changed.

## Behavior

Unchanged. The five replaced expressions were byte-identical to each other — same `''` default, same `.lower()`, same `== 'true'` comparison — and the new helper keeps all three. Every original check already sat inside a function body, so evaluation still happens at call time, not at import time; `main_http.py:50` setting the env var before importing `server` remains correct.

The rename from `_is_auth_enabled` to `is_auth_enabled` drops the underscore because the function is now imported by six modules. A leading underscore on a symbol with cross-module callers was the pre-existing smell that `tools/webftp_tools.py` was already violating.

## Testing

- `uv run --extra dev python -m pytest tests -q`: 949 passed, 2 failed. Both failures (`tests/test_basic.py::test_import_main`, `tests/test_simple_integration.py::TestMCPServerConfiguration::test_main_entry_point`) reproduce on a clean tree — verified with `git stash` — and are caused by the local checkout directory being named `main`, so `import main` resolves to the package instead of `main.py`. Not related to this change.
- `ruff check .` and `ruff format --check .`: clean.
- Tests covering both branches of the flag (`tests/test_http_client.py`, `tests/test_health.py`, `tests/test_input_validation.py`, `tests/test_oauth.py`, `tests/test_mfa_precheck.py`) set `ALPACON_MCP_AUTH_ENABLED` via `patch.dict(os.environ, ...)` rather than patching a function, so they exercise the new helper unmodified.

## Checklist

- [x] No behavior change intended, and the replaced expressions were verified identical before removal
- [x] Tests pass locally (2 pre-existing failures documented above)
- [x] Lint and format clean
- [x] No new dependencies

### DIFF
--- a/tools/workspace_tools.py
+++ b/tools/workspace_tools.py
@@ -5,7 +5,12 @@ from typing import Any
 from mcp.types import ToolAnnotations
 
 from server import mcp
-from utils.common import error_response, success_response, unwrap_http_result
+from utils.common import (
+    error_response,
+    is_auth_enabled,
+    success_response,
+    unwrap_http_result,
+)
 from utils.decorators import mcp_tool_handler, require_jwt_auth
 from utils.http_client import http_client
 from utils.token_manager import TokenManager
@@ -115,10 +120,10 @@ async def list_workspaces(region: str = '') -> dict[str, Any]:
     Returns:
         Workspaces list response
     """
-    from utils.decorators import _get_jwt_token, _get_jwt_workspaces, _is_auth_enabled
+    from utils.decorators import _get_jwt_token, _get_jwt_workspaces
 
     # Use explicit transport mode check, not JWT presence
-    if _is_auth_enabled():
+    if is_auth_enabled():
         jwt_token = _get_jwt_token()
         if not jwt_token:
             return error_response(
--- a/server.py
+++ b/server.py
@@ -8,6 +8,7 @@ from typing import Literal
 
 from mcp.server.fastmcp import FastMCP
 
+from utils.common import is_auth_enabled
 from utils.logger import get_logger
 
 logger = get_logger('server')
@@ -74,7 +75,7 @@ def _create_mcp_server() -> FastMCP:
     creates the server with Auth0 JWT authentication for HTTP transport.
     Otherwise creates a standard server for stdio/SSE transport.
     """
-    auth_enabled = os.getenv('ALPACON_MCP_AUTH_ENABLED', '').lower() == 'true'
+    auth_enabled = is_auth_enabled()
 
     if auth_enabled:
         from mcp.server.auth.settings import AuthSettings
@@ -254,11 +255,6 @@ def _modules_to_load(toolsets: str | None, remote_mode: bool) -> set[str]:
     return enabled | ALWAYS_ON_MODULES
 
 
-def _is_remote_mode() -> bool:
-    """Check if running in remote (streamable-http) mode with JWT auth."""
-    return os.getenv('ALPACON_MCP_AUTH_ENABLED', '').lower() == 'true'
-
-
 def _install_upstream_auth_middleware():
     """Override run_streamable_http_async to wrap app with auth error middleware.
 
@@ -346,7 +342,7 @@ def run(
     else:
         logger.info('No config file specified, using default config discovery')
 
-    remote_mode = _is_remote_mode()
+    remote_mode = is_auth_enabled()
 
     if remote_mode:
         # Remote (streamable-http) mode: register OAuth routes
--- a/tests/test_webftp_tools.py
+++ b/tests/test_webftp_tools.py
@@ -1255,9 +1255,9 @@ class TestRemoteModeUnsupported:
         """webftp_upload_file returns remote_mode_unsupported error in remote mode.
 
         The decorator's auth check is bypassed (auth disabled) so the function
-        body's _is_remote_mode check is what we're exercising here.
+        body's is_auth_enabled check is what we're exercising here.
         """
-        with patch('tools.webftp_tools._is_auth_enabled', return_value=True):
+        with patch('tools.webftp_tools.is_auth_enabled', return_value=True):
             result = await webftp_upload_file(
                 server_id=self.SERVER_ID,
                 local_file_path='/local/test.txt',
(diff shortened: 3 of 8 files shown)


tools/workspace_tools.py
```diff
@@ -115,10 +120,10 @@ async def list_workspaces(region: str = '') -> dict[str, Any]:
     Returns:
         Workspaces list response
     """
-    from utils.decorators import _get_jwt_token, _get_jwt_workspaces, _is_auth_enabled
+    from utils.decorators import _get_jwt_token, _get_jwt_workspaces
```
junho226, line 123: Suggestion: `_is_auth_enabled` is gone from this line, but `_get_jwt_token` / `_get_jwt_workspaces` are still module-private symbols imported across a module boundary — the same smell this PR fixes. Also, `utils.decorators` is already imported at module level (line 14), so there is no cycle forcing this function-local import. A follow-up could promote these two to public names (or move the JWT-context helpers next to `is_auth_enabled`) and lift the import to the top of the file. Non-blocking.

junho226 (approved):
## Summary

Consolidates the five copy-pasted `ALPACON_MCP_AUTH_ENABLED` checks (plus two identically-bodied private helpers and one cross-module private import) into a single `utils.common.is_auth_enabled()`. Pure refactor, no behavior change.

## Verification

I verified the PR's claims independently in a local worktree:

- `grep` confirms `utils/common.py:114` is now the only reader of the env var — remaining hits are the setter in `main_http.py`, docstrings, and tests that set the var via `patch.dict`.
- No leftover `_is_remote_mode` / `_is_auth_enabled` definitions or imports anywhere.
- `ruff check .` and `ruff format --check .` both clean (also confirms no unused `os` imports were left behind).
- `uv run --extra dev python -m pytest tests -q`: **951 passed** — including the 2 tests that failed on the author's checkout, which confirms the diagnosis that those failures come from the checkout directory being named `main`, not from this change.
- The reorder commit (74e90c0) is verified pure code motion: the non-blank line multiset of `utils/common.py` is identical before and after.
- Import graph checked: `utils/common.py` stays below every consumer; the stated reason for not hosting the helper in `utils/decorators.py` (cycle via `utils/http_client.py`) holds.

## Suggestions (1, non-blocking)

- `tools/workspace_tools.py:123` — see inline comment: the sibling private imports `_get_jwt_token` / `_get_jwt_workspaces` are the same smell this PR fixes, a natural follow-up.

## Highlights

- Every verifiable claim in the PR body reproduced exactly, including the test-failure diagnosis.
- Test patch targets correctly moved to the consumer module (`tools.webftp_tools.is_auth_enabled`), matching how `from`-imports bind.
- The new docstring merges both removed helpers' perspectives (cause: env var, effect: remote mode).

**Verdict**: APPROVED

- [x] Tests included and passing (951 passed locally)
- [x] No security vulnerabilities
- [x] Documentation updated where needed (docstrings)

tools/workspace_tools.py
jisung-02: Agreed on both points, and I confirmed them: `utils.decorators` is already imported at module level (line 14), so the function-local import here is not breaking a cycle, and `tools/workspace_tools.py` is the only cross-module user of `_get_jwt_token` / `_get_jwt_workspaces` (the remaining hits are inside `utils/decorators.py` itself and `patch(...)` targets in tests).

Splitting this into a follow-up PR rather than folding it in here: promoting the two names touches the four test modules that patch `utils.decorators._get_jwt_token`, and where the JWT-context helpers should live (public in `utils/decorators.py` vs. moved next to `is_auth_enabled` in `utils/common.py`) is its own call. Keeping this PR to the single `ALPACON_MCP_AUTH_ENABLED` consolidation.

PR #142, opened 2026-07-28
refactor(utils): consolidate retry backoff and UUID list validation

jisung-02:
## Background

Two duplication sites in `utils/`, plus one layout problem that made the first one hard to see.

In `AlpaconHTTPClient.request()`, the 5xx, timeout, and network handlers each repeated the same four lines: bump `retry_count`, log the attempt, sleep, double the delay capped at `max_retry_delay`. The three copies were not byte-identical — the log messages differed and each handler ended differently — which is what let a real defect hide in the 5xx arm.

In `utils/decorators.py`, the `server_ids` and `servers` arguments were validated by two blocks that differed only in variable names: same `isinstance` list guard, same comprehension over `validate_server_id_format`, same two error strings.

Separately, `AlpaconHTTPClient` interleaved public and private members. `__aenter__` and `__aexit__` sat at the very end of the class, roughly 250 lines below the `_close_client` they call, and the four private cache helpers split `close()` from `get_base_url()`.

A third pass over the same two files turned up inline literals of the same kind: values repeated across sites, or rebuilt on every call. `_is_cacheable()` reconstructed a seven-element list of endpoint prefixes on each of the up to three calls per request, and `with_logging` rebuilt its redaction list per tool call.

## What changed

Ten commits, one concern each.

`ba7bb63` adds a local `backoff()` closure inside `request()` that owns the count/log/sleep/double sequence and returns `False` once the attempts are spent. Each handler is left with only its own exhaustion arm.

`8788207` extracts `_validate_uuid_list(field, value)` and drives both arguments through a loop over `('server_ids', 'servers')`.

`0859978` moves the exhaustion check ahead of the retry log inside `backoff()`, so the last failed attempt no longer announces a retry that will not happen.

`0960d49` regroups the class members: lifecycle (`__init__`, the async context manager pair, `close`, `_close_client`), then properties, then public API, then private helpers last. The `__del__ removed` note moves with the lifecycle block.

`29700ae` names the inline literals. In `utils/http_client.py`: the seven `error` codes, the backoff multiplier, and `_CACHEABLE_ENDPOINTS`, now a tuple so it feeds `str.startswith` directly and the `any()` generator goes away. In `utils/decorators.py`: the two UUID validation messages, the field tuple driving the loop, the sentence shared by both "multiple regions available" errors, and the redaction keys, now a `frozenset` matching the `_WORK_SESSION_GATE_CODES` shape in `utils/common.py`. `ba16afb` walks three of these back.

`24d23d2` trims the comments in these two files, keeping only the ones stating something the code cannot. Mostly the ones the earlier commits added, plus one pre-existing block: the eight lines in `_handle_upstream_401` explaining the token-hash dict opened by restating the docstring three lines above them, so they go down to two.

`4e40dcc` adds a test for the 5xx exhaustion response, which the three existing tests only assert the `error` key of. It drives the real exhaustion path and pins the status code, the domain `_detect_error_domain` resolves, and the exact hints `enrich_error_response` attaches.

`ba16afb` reverses the three constants from `29700ae` that bought nothing. `_BACKOFF_MULTIPLIER` had one call site 167 lines from its definition and was the wrong shape besides — the other three retry knobs are instance attributes, so the multiplier joins them as `self.backoff_multiplier`. The two UUID messages were already deduped by `8788207`, whose `_validate_uuid_list` is their only reader, so hoisting them moved the text ~210 lines from its sole use site without removing a duplicate; they go back inline, and `_UUID_LIST_FIELDS` follows. The two `f'{_SPECIFY_REGION_HINT}'` sites become implicit concatenation of the plain name. What survives from `29700ae` is what earns it: `_CACHEABLE_ENDPOINTS` and `_SENSITIVE_LOG_KEYS` remove a per-call rebuild, the seven `_ERR_*` are a contract catalog pinned by 17 test assertions, and `_SPECIFY_REGION_HINT` itself has two call sites.

`283c326` fixes the test `4e40dcc` added. It called `enrich_error_response` without `tool_name`, but the live path in `with_error_handling` always passes it and `_detect_error_domain` reads `tool_name` ahead of the message, so the webftp and command tools resolve to `file` and `command` rather than the `server` the message alone implies. It changes nothing today — `_HINT_REGISTRY` has only a `(500, 'general')` entry, so every domain falls through to the same hints — but adding a `(500, 'server')` or `(500, 'file')` entry would change what those tools receive while a test pinning the no-argument call kept passing. The test now walks three tool names covering the three domains and asserts both the domain each resolves to and that the hints still converge.

`2456f32` drops the backoff sleep from the four tests that exhaust the retries. `test_get_500_error`, the domain test above, `test_network_timeout`, and `test_connection_error` were 12s of an 18.5s suite, and none of them tests the timing. A `no_retry_delay` fixture zeroes `retry_delay` via `monkeypatch` — a direct write would leak, since `http_client` is a module-level singleton.

## Behavior

Two deliberate fixes, one added field with a downstream effect of its own, otherwise unchanged.

The 5xx branch had no exhaustion arm. When retries ran out it fell out of the loop into the tail fallback, which logged `Unexpected fallback` even though three tests assert exactly that path as the normal outcome (`tests/test_http_client.py:148`, `tests/integration/test_http_client_integration.py:40` and `:171`). It now returns in place, and the tail return — reachable only when `max_retries <= 0` — has its comment and log corrected to say so.

That in-place 5xx response also carries `status_code`, which the old tail response dropped. The `error` key stays `'Max retries exceeded'` because the three tests above assert on it.

The added `status_code` is not inert, and neither is the accompanying message change from `'Failed after N attempts'` to `'Server error after N attempts'`. Both are read downstream: `utils/common.py:292` lifts `status_code` into the error `unwrap_http_result` returns, and `utils/recovery_hints.py:220` reads it while `utils/decorators.py:407` runs `enrich_error_response` over every decorated tool result. The message is also an input to `_detect_error_domain`, which keys on the substring `server`. So a 5xx exhaustion that previously enriched to `['error', 'message']` with no hints now enriches to `['error', 'message', 'recovery_hints', 'status_code']` with two server hints. This is a third deliberate delta, not a payload detail: the agent now gets actionable guidance where it previously got a bare string. `tests/test_http_client.py:161` pins it, so whoever next edits `message` finds out they moved the domain detection.

The second fix is log-only. `backoff()` counted the attempt, logged `retrying (3/3) in 4.0s`, and only then checked whether the attempts were spent, so the final failure claimed a retry that never followed. The check now comes first and `backoff()` returns without logging; the three callers already log the exhaustion with the same reason, so nothing is lost. All three pre-refactor copies had the same misordering, so this is a pre-existing log defect the consolidation made visible rather than one it introduced. No return value or error payload changes.

The UUID validation loop preserves the original field order, so a call with both arguments invalid still reports `server_ids` first, and the error payload is unchanged.

`29700ae`, `24d23d2`, and `ba16afb` change no behavior. Each constant was checked to equal the literal it replaced byte for byte, and `_is_cacheable` was compared against the old `any(startswith)` form across 32 method/endpoint cases, including case differences, a missing leading slash, and a mid-string match. `24d23d2` touches only comment and docstring lines. The literals `ba16afb` puts back were checked against the parent revision the same way, and `2` becomes `2.0` only to match the float retry knobs it now sits beside — the product was already a float.

One comment introduced by `29700ae` was wrong and `24d23d2` corrects it. It described the error codes as values "callers branch on"; no production consumer does. `utils/common.py:288`, `utils/recovery_hints.py:212`, and `tools/server_tools.py:67` all test only for the presence of the `error` key. What pins the values is 17 verbatim test assertions, which is what the comment says now.

`0960d49` is pure code motion. With blank lines dropped, the 581-line multiset of `utils/http_client.py` is identical before and after — verified by comparing the two revisions directly, the same check documented in `74e90c0`. No signature, body, or import changed.

## Testing

- `uv run --extra dev python -m pytest tests -q`: 954 passed in 6.4s, down from 18.5s.
- `tests/test_http_client.py:161` is new. It drives the real 5xx exhaustion path once per domain and asserts, for each, that `status_code` is on the response, that `_detect_error_domain` resolves the domain the live path would, and that the two hints `enrich_error_response` attaches are the same across all three.
- `uv run --extra dev ruff check .`: clean. `ruff format --check .`: 82 files already formatted.
- `tests/test_input_validation.py:244-340` already covered `server_ids` (invalid, mixed, `None`, non-list) and `servers` (invalid, mixed), so the extraction is exercised unmodified.
- `tests/test_shutdown.py:59-70` tests `_close_client()` as a backward-compat alias, which is why it stays next to `close()` rather than moving to the private-helper group.
- The forward hazard the domain test guards was checked rather than assumed: injecting a `(500, 'file')` entry into `_HINT_REGISTRY` and running that test alone fails it. Without `tool_name` it kept passing.

Two literals were deliberately left inline. The PyJWT decode-options dict in `_decode_jwt_claims` looks constant, but PyJWT 2.13.0's `_merge_options` writes `verify_nbf`, `verify_iat`, `verify_sub`, and `verify_jti` back into the caller's dict when `verify_signature` is false, so a module-level constant would be mutated by the library on every call. The `401` and `500` status codes stay as integers because `http.HTTPStatus` would put an `IntEnum` into the response dict, which the log f-strings render as `<HTTPStatus.UNAUTHORIZED: 401>`.

## Checklist

- [x] The three backoff copies were compared before collapsing them; the differences are called out above rather than assumed away
- [x] The 5xx behavior change is intentional and named, not incidental, down to what the added `status_code` and the new message do to the enriched response
- [x] The retry-log fix was checked against the rest of the codebase; no other log announces an action before the guard that decides it
- [x] The constants preserve their literals byte for byte, verified against the parent revision
- [x] One inaccurate comment was found and corrected rather than carried forward
- [x] Three constants from `29700ae` were walked back once review showed they bought nothing; the ones that survive each have a stated reason
- [x] The new test mirrors the live `enrich_error_response` call rather than a shape no caller uses
- [x] Code motion verified pure by line-multiset comparison
- [x] Tests pass locally, lint and format clean
- [x] No new dependencies

### DIFF
--- a/tests/test_http_client.py
+++ b/tests/test_http_client.py
@@ -11,6 +11,7 @@ import httpx
 import pytest
 
 from utils.http_client import AlpaconHTTPClient, http_client
+from utils.recovery_hints import _detect_error_domain, enrich_error_response
 
 
 @pytest.fixture
@@ -137,6 +138,36 @@ class TestHTTPClientGet:
         assert result['error'] == 'Max retries exceeded'
         # Should have retried 3 times
         assert mock_httpx_client.request.call_count == 3
+
+    @pytest.mark.asyncio
+    async def test_get_500_exhaustion_enriches_with_server_hints(
+        self, mock_httpx_client
+    ):
+        """5xx exhaustion carries status_code and a 'server'-domain message.
+
+        Both feed enrich_error_response, which every decorated tool runs over
+        its result, so changing either moves the hints the caller receives.
+        """
+        mock_response = create_mock_response(status_code=500)
+        mock_httpx_client.request.return_value = mock_response
+
+        result = await http_client.get(
+            region='ap1',
+            workspace='testworkspace',
+            endpoint='/api/error/',
+            token='test-token',
+        )
+
+        assert result['status_code'] == 500
+        assert (
+            _detect_error_domain(result['status_code'], result['message']) == 'server'
+        )
+
+        enriched = enrich_error_response(dict(result))
+        assert enriched['recovery_hints'] == [
+            'The server encountered an internal error. Try again in a moment.',
+            'If the issue persists, check server health.',
+        ]
 
 
 class TestHTTPClientPost:
--- a/tests/test_http_client.py
+++ b/tests/test_http_client.py
@@ -41,6 +41,16 @@ def mock_httpx_client():
             delattr(http_client, '_disable_pooling')
 
 
+@pytest.fixture
+def no_retry_delay(monkeypatch):
+    """Drop the backoff sleep for tests that exhaust the retries.
+
+    monkeypatch, not assignment: http_client is a module-level singleton, so
+    the value would leak into the rest of the suite.
+    """
+    monkeypatch.setattr(http_client, 'retry_delay', 0)
+
+
 def create_mock_response(status_code=200, json_data=None, text_data='', headers=None):
     """Create a properly configured mock response."""
     # Use MagicMock instead of AsyncMock for response to avoid coroutine issues
@@ -123,7 +133,7 @@ class TestHTTPClientGet:
         assert result['status_code'] == 404
 
     @pytest.mark.asyncio
-    async def test_get_500_error(self, mock_httpx_client):
+    async def test_get_500_error(self, mock_httpx_client, no_retry_delay):
         """Test GET request with 500 error (should retry and then fail)."""
         mock_response = create_mock_response(status_code=500)
         mock_httpx_client.request.return_value = mock_response
@@ -139,14 +149,21 @@ class TestHTTPClientGet:
         # Should have retried 3 times
         assert mock_httpx_client.request.call_count == 3
 
+    @pytest.mark.parametrize(
+        ('tool_name', 'domain'),
+        [
+            ('get_server', 'server'),
+            ('execute_command', 'command'),
+            ('webftp_upload_file', 'file'),
+        ],
+    )
     @pytest.mark.asyncio
     async def test_get_500_exhaustion_enriches_across_tool_domains(
-        self, mock_httpx_client
+        self, mock_httpx_client, no_retry_delay, tool_name, domain
     ):
         """tool_name mirrors the live enrich call and outranks the message, so the
         domain varies by caller; 500 has only a (500, 'general') entry so the hints
-        converge, and a domain-specific one would split them here. One request for
-        all three — exhausting the retries really sleeps.
+        converge, and a domain-specific one would split them here.
         """
         mock_response = create_mock_response(status_code=500)
         mock_httpx_client.request.return_value = mock_response
@@ -159,23 +176,16 @@ class TestHTTPClientGet:
         )
 
         assert result['status_code'] == 500
+        assert (
+            _detect_error_domain(result['status_code'], result['message'], tool_name)
+            == domain
+        )
 
-        for tool_name, domain in (
-            ('get_server', 'server'),
-            ('execute_command', 'command'),
-            ('webftp_upload_file', 'file'),
-        ):
-            assert (
-                _detect_error_domain(
-                    result['status_code'], result['message'], tool_name
-                )
-                == domain
-            )
-            enriched = enrich_error_response(dict(result), tool_name=tool_name)
-            assert enriched['recovery_hints'] == [
-                'The server encountered an internal error. Try again in a moment.',
-                'If the issue persists, check server health.',
-            ]
+        enriched = enrich_error_response(dict(result), tool_name=tool_name)
+        assert enriched['recovery_hints'] == [
+            'The server encountered an internal error. Try again in a moment.',
+            'If the issue persists, check server health.',
+        ]
 
 
 class TestHTTPClientPost:
@@ -374,7 +384,7 @@ class TestErrorHandling:
     """Test error handling scenarios."""
 
     @pytest.mark.asyncio
-    async def test_network_timeout(self, mock_httpx_client):
+    async def test_network_timeout(self, mock_httpx_client, no_retry_delay):
         """Test network timeout handling."""
         mock_httpx_client.request.side_effect = httpx.TimeoutException(
             'Request timeout'
@@ -390,7 +400,7 @@ class TestErrorHandling:
         assert result['error'] == 'Timeout'
 
     @pytest.mark.asyncio
-    async def test_connection_error(self, mock_httpx_client):
+    async def test_connection_error(self, mock_httpx_client, no_retry_delay):
         """Test connection error handling."""
         mock_httpx_client.request.side_effect = httpx.ConnectError('Connection failed')
 
--- a/tests/test_http_client.py
+++ b/tests/test_http_client.py
@@ -140,13 +140,13 @@ class TestHTTPClientGet:
         assert mock_httpx_client.request.call_count == 3
 
     @pytest.mark.asyncio
-    async def test_get_500_exhaustion_enriches_with_server_hints(
+    async def test_get_500_exhaustion_enriches_across_tool_domains(
         self, mock_httpx_client
     ):
-        """5xx exhaustion carries status_code and a 'server'-domain message.
-
-        Both feed enrich_error_response, which every decorated tool runs over
-        its result, so changing either moves the hints the caller receives.
+        """tool_name mirrors the live enrich call and outranks the message, so the
+        domain varies by caller; 500 has only a (500, 'general') entry so the hints
+        converge, and a domain-specific one would split them here. One request for
+        all three — exhausting the retries really sleeps.
         """
         mock_response = create_mock_response(status_code=500)
         mock_httpx_client.request.return_value = mock_response
@@ -159,15 +159,23 @@ class TestHTTPClientGet:
         )
 
         assert result['status_code'] == 500
-        assert (
-            _detect_error_domain(result['status_code'], result['message']) == 'server'
-        )
 
-        enriched = enrich_error_response(dict(result))
-        assert enriched['recovery_hints'] == [
-            'The server encountered an internal error. Try again in a moment.',
-            'If the issue persists, check server health.',
-        ]
+        for tool_name, domain in (
+            ('get_server', 'server'),
+            ('execute_command', 'command'),
+            ('webftp_upload_file', 'file'),
+        ):
+            assert (
+                _detect_error_domain(
+                    result['status_code'], result['message'], tool_name
+                )
+                == domain
+            )
+            enriched = enrich_error_response(dict(result), tool_name=tool_name)
+            assert enriched['recovery_hints'] == [
+                'The server encountered an internal error. Try again in a moment.',
+                'If the issue persists, check server health.',
+            ]
 
 
 class TestHTTPClientPost:
--- a/utils/decorators.py
+++ b/utils/decorators.py
@@ -28,16 +28,13 @@ from utils.logger import get_logger
 
 logger = get_logger('decorators')
 
-# Arguments carrying a list of server UUIDs. Order is the reporting order when
-# more than one is invalid.
+# Order is the reporting order when more than one is invalid.
 _UUID_LIST_FIELDS = ('server_ids', 'servers')
 _UUID_LIST_TYPE_MSG = 'Must be a list of server UUIDs.'
 _UUID_LIST_ITEM_MSG = 'Each server ID must be in UUID format. (e.g., 550e8400-e29b-41d4-a716-446655440000)'
 
-# Trailing hint shared by the two "multiple regions available" messages.
 _SPECIFY_REGION_HINT = 'Please specify a region parameter.'
 
-# Argument names redacted from tool-entry logs.
 _SENSITIVE_LOG_KEYS = frozenset({'_token', 'password', 'secret', 'key'})
 
 
@@ -238,7 +235,7 @@ async def _check_mfa_requirement(
 
 
 def _validate_uuid_list(field: str, value: Any) -> dict[str, Any] | None:
-    """Validate a list of server UUIDs. Returns an error response, or None if valid."""
+    """Returns an error response, or None if valid."""
     if not isinstance(value, list):
         return format_validation_error(
             field,
@@ -327,8 +324,7 @@ def with_token_validation(func: Callable) -> Callable:
         if server_id is not None and not validate_server_id_format(server_id):
             return format_validation_error('server_id', server_id)
 
-        # Validate UUID list arguments if present
-        # ('servers' carries server UUIDs sent in request bodies)
+        # 'servers' carries server UUIDs sent in request bodies.
         for field in _UUID_LIST_FIELDS:
             value = arguments.get(field)
             if value is not None:
--- a/utils/decorators.py
+++ b/utils/decorators.py
@@ -225,6 +225,25 @@ async def _check_mfa_requirement(
         logger.debug('MFA pre-check failed (non-fatal): %s', e)
 
 
+def _validate_uuid_list(field: str, value: Any) -> dict[str, Any] | None:
+    """Validate a list of server UUIDs. Returns an error response, or None if valid."""
+    if not isinstance(value, list):
+        return format_validation_error(
+            field,
+            value,
+            'Must be a list of server UUIDs.',
+        )
+
+    invalid = [item for item in value if not validate_server_id_format(item)]
+    if invalid:
+        return format_validation_error(
+            field,
+            invalid,
+            'Each server ID must be in UUID format. (e.g., 550e8400-e29b-41d4-a716-446655440000)',
+        )
+    return None
+
+
 def with_token_validation(func: Callable) -> Callable:
     """Decorator to add automatic token validation to MCP tools.
 
@@ -296,43 +315,14 @@ def with_token_validation(func: Callable) -> Callable:
         if server_id is not None and not validate_server_id_format(server_id):
             return format_validation_error('server_id', server_id)
 
-        # Validate server_ids list if present
-        server_ids = arguments.get('server_ids')
-        if server_ids is not None:
-            if not isinstance(server_ids, list):
-                return format_validation_error(
-                    'server_ids',
-                    server_ids,
-                    'Must be a list of server UUIDs.',
-                )
-            invalid_ids = [
-                sid for sid in server_ids if not validate_server_id_format(sid)
-            ]
-            if invalid_ids:
-                return format_validation_error(
-                    'server_ids',
-                    invalid_ids,
-                    'Each server ID must be in UUID format. (e.g., 550e8400-e29b-41d4-a716-446655440000)',
-                )
-
-        # Validate servers list if present (server UUIDs sent in request bodies)
-        servers = arguments.get('servers')
-        if servers is not None:
-            if not isinstance(servers, list):
-                return format_validation_error(
-                    'servers',
-                    servers,
-                    'Must be a list of server UUIDs.',
-                )
-            invalid_servers = [
-                sid for sid in servers if not validate_server_id_format(sid)
-            ]
-            if invalid_servers:
-                return format_validation_error(
-                    'servers',
-                    invalid_servers,
-                    'Each server ID must be in UUID format. (e.g., 550e8400-e29b-41d4-a716-446655440000)',
-                )
+        # Validate UUID list arguments if present
+        # ('servers' carries server UUIDs sent in request bodies)
+        for field in ('server_ids', 'servers'):
+            value = arguments.get(field)
+            if value is not None:
+                validation_error = _validate_uuid_list(field, value)
+                if validation_error:
+                    return validation_error
 
         # session_id is interpolated into URL paths, so reject non-UUID values that could retarget the request.
         session_id = arguments.get('session_id')
--- a/utils/decorators.py
+++ b/utils/decorators.py
@@ -28,6 +28,18 @@ from utils.logger import get_logger
 
 logger = get_logger('decorators')
 
+# Arguments carrying a list of server UUIDs. Order is the reporting order when
+# more than one is invalid.
+_UUID_LIST_FIELDS = ('server_ids', 'servers')
+_UUID_LIST_TYPE_MSG = 'Must be a list of server UUIDs.'
+_UUID_LIST_ITEM_MSG = 'Each server ID must be in UUID format. (e.g., 550e8400-e29b-41d4-a716-446655440000)'
+
+# Trailing hint shared by the two "multiple regions available" messages.
+_SPECIFY_REGION_HINT = 'Please specify a region parameter.'
+
+# Argument names redacted from tool-entry logs.
+_SENSITIVE_LOG_KEYS = frozenset({'_token', 'password', 'secret', 'key'})
+
 
 def _get_jwt_token() -> str | None:
     """Get JWT token from FastMCP auth context if available.
@@ -133,7 +145,7 @@ def _resolve_region_jwt(
     if available_regions:
         return None, (
             f'Multiple regions available in token: {", ".join(available_regions)}. '
-            f'Please specify a region parameter.'
+            f'{_SPECIFY_REGION_HINT}'
         )
     return None, 'No regions found in JWT token.'
 
@@ -161,7 +173,7 @@ def _resolve_region_local(workspace: str | None) -> tuple[str | None, str | None
     if available_regions:
         return None, (
             f'Multiple regions available: {", ".join(sorted(available_regions))}. '
-            f'Please specify a region parameter.'
+            f'{_SPECIFY_REGION_HINT}'
         )
     return None, 'No regions configured. Please run setup first.'
 
@@ -231,7 +243,7 @@ def _validate_uuid_list(field: str, value: Any) -> dict[str, Any] | None:
         return format_validation_error(
             field,
             value,
-            'Must be a list of server UUIDs.',
+            _UUID_LIST_TYPE_MSG,
         )
 
     invalid = [item for item in value if not validate_server_id_format(item)]
@@ -239,7 +251,7 @@ def _validate_uuid_list(field: str, value: Any) -> dict[str, Any] | None:
         return format_validation_error(
             field,
             invalid,
-            'Each server ID must be in UUID format. (e.g., 550e8400-e29b-41d4-a716-446655440000)',
+            _UUID_LIST_ITEM_MSG,
         )
     return None
 
@@ -317,7 +329,7 @@ def with_token_validation(func: Callable) -> Callable:
 
         # Validate UUID list arguments if present
         # ('servers' carries server UUIDs sent in request bodies)
-        for field in ('server_ids', 'servers'):
+        for field in _UUID_LIST_FIELDS:
             value = arguments.get(field)
             if value is not None:
                 validation_error = _validate_uuid_list(field, value)
@@ -455,11 +467,7 @@ def with_logging(func: Callable) -> Callable:
         arguments = bound_args.arguments
 
         # Create log-safe arguments (exclude sensitive data)
-        log_args = {
-            k: v
-            for k, v in arguments.items()
-            if k not in ['_token', 'password', 'secret', 'key']
-        }
+        log_args = {k: v for k, v in arguments.items() if k not in _SENSITIVE_LOG_KEYS}
 
         # Log function entry
         logger.info(f'{func_name} called with: {log_args}')
--- a/utils/decorators.py
+++ b/utils/decorators.py
@@ -28,11 +28,6 @@ from utils.logger import get_logger
 
 logger = get_logger('decorators')
 
-# Order is the reporting order when more than one is invalid.
-_UUID_LIST_FIELDS = ('server_ids', 'servers')
-_UUID_LIST_TYPE_MSG = 'Must be a list of server UUIDs.'
-_UUID_LIST_ITEM_MSG = 'Each server ID must be in UUID format. (e.g., 550e8400-e29b-41d4-a716-446655440000)'
-
 _SPECIFY_REGION_HINT = 'Please specify a region parameter.'
 
 _SENSITIVE_LOG_KEYS = frozenset({'_token', 'password', 'secret', 'key'})
@@ -142,7 +137,7 @@ def _resolve_region_jwt(
     if available_regions:
         return None, (
             f'Multiple regions available in token: {", ".join(available_regions)}. '
-            f'{_SPECIFY_REGION_HINT}'
+            + _SPECIFY_REGION_HINT
         )
     return None, 'No regions found in JWT token.'
 
@@ -170,7 +165,7 @@ def _resolve_region_local(workspace: str | None) -> tuple[str | None, str | None
     if available_regions:
         return None, (
             f'Multiple regions available: {", ".join(sorted(available_regions))}. '
-            f'{_SPECIFY_REGION_HINT}'
+            + _SPECIFY_REGION_HINT
         )
     return None, 'No regions configured. Please run setup first.'
 
@@ -240,7 +235,7 @@ def _validate_uuid_list(field: str, value: Any) -> dict[str, Any] | None:
         return format_validation_error(
             field,
             value,
-            _UUID_LIST_TYPE_MSG,
+            'Must be a list of server UUIDs.',
         )
 
     invalid = [item for item in value if not validate_server_id_format(item)]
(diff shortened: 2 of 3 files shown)


utils/http_client.py
```diff
@@ -322,6 +156,19 @@ async def request(
...
+            nonlocal retry_count, retry_delay
+            retry_count += 1
+            logger.warning(
+                f'{reason}, retrying ({retry_count}/{self.max_retries}) in {retry_delay}s'
+            )
+            if retry_count >= self.max_retries:
+                return False
+            await asyncio.sleep(retry_delay)
+            retry_delay = min(retry_delay * 2, self.max_retry_delay)
+            return True
```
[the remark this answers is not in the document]
jisung-02: Fixed in 0859978. The exhaustion check now runs before the log, so the final failed attempt no longer announces a retry that will not happen.

The reason string is not lost on the exhaustion path: each of the three callers already logs it with its own error response (`Server error after all retries`, `Request timeout after all retries`, `Network error after all retries`), so `backoff()` returns silently rather than emitting a second "exhausted" line that would duplicate them.

Worth noting this is pre-existing, not introduced by the consolidation — all three original copies logged before their `if retry_count < self.max_retries:` guard. The consolidation is what made the single shared misordering visible.

Checked the rest of the codebase for the same class of issue (a log describing an action emitted before the guard that decides whether it happens). This was the only occurrence: the retry loop in `request()` is the only retry loop, and the cooldown logs in `utils/auth_error_middleware.py` are all emitted inside their respective branches. No test asserted on the retry log text. 951 tests pass, ruff check and format clean.

utils/http_client.py
```diff
@@ -392,21 +262,23 @@ async def request(
...
-                    )
-                    if retry_count < self.max_retries:
-                        await asyncio.sleep(retry_delay)
-                        retry_delay = min(retry_delay * 2, self.max_retry_delay)
+                    if await backoff('Server error'):
                         continue
+
+                    error_response = {
+                        'error': _ERR_MAX_RETRIES,
+                        'status_code': e.response.status_code,
```
junho226, line 271: Warning: this added `status_code` is not inert.

`utils/common.py:292` lifts it into `error_kwargs`, and `utils/recovery_hints.py:220` reads it on the live path through `utils/decorators.py:407` (`enrich_error_response` over every decorated tool result). Together with the `message` change (`Failed after N attempts` → `Server error after N attempts`, which also flips `_detect_error_domain` on the substring `server`), a 5xx exhaustion now returns two keys it never returned before:

```
--- OLD ---
  keys : [error, message]
  hints: None
--- NEW ---
  keys : [error, message, recovery_hints, status_code]
  hints: [The server encountered an internal error. Try again in a moment., If the issue persists, check server health.]
```

This reads as an improvement, and nothing regressed because the three tests on this path assert only on `error`. But the body names the field without naming the consequence, and this PR's own standard is that every delta is named. Worth a line in the description plus a test pinning `recovery_hints` here, so the next person to touch `message` learns they moved the domain detection.

utils/http_client.py
```diff
@@ -13,6 +13,28 @@
...
+# Pinned values: tests assert these verbatim; callers only check that 'error' exists.
+_ERR_HTTP = 'HTTP Error'
+_ERR_MAX_RETRIES = 'Max retries exceeded'
+_ERR_MFA_REQUIRED = 'MFA Required'
+_ERR_REQUEST = 'Request Error'
+_ERR_REQUEST_EXCEPTION = 'Request Exception'
+_ERR_TIMEOUT = 'Timeout'
+_ERR_UNEXPECTED = 'Unexpected Error'
+
+_BACKOFF_MULTIPLIER = 2
```
junho226: Suggestion: single call site, and the literal reads better than the name.

`retry_delay = min(retry_delay * 2, self.max_retry_delay)` is self-documenting next to `max_retry_delay`; the constant adds a lookup 167 lines away to learn that doubling is doubling. Of everything `29700ae` hoisted, only `_SPECIFY_REGION_HINT` has two call sites — `_CACHEABLE_ENDPOINTS` and `_SENSITIVE_LOG_KEYS` earn their place by removing a per-call rebuild, and the `_ERR_*` set is defensible as a test-pinned contract catalog. This one buys neither.

```suggestion
```

utils/decorators.py
```diff
@@ -28,6 +28,15 @@
 
 logger = get_logger('decorators')
 
+# Order is the reporting order when more than one is invalid.
+_UUID_LIST_FIELDS = ('server_ids', 'servers')
+_UUID_LIST_TYPE_MSG = 'Must be a list of server UUIDs.'
+_UUID_LIST_ITEM_MSG = 'Each server ID must be in UUID format. (e.g., 550e8400-e29b-41d4-a716-446655440000)'
```
junho226: Suggestion: these two are single-use after the extraction they were hoisted out of.

`8788207` already deduped them by creating `_validate_uuid_list`, which is now their only reader. Hoisting them in `29700ae` moves the text ~210 lines from its sole use site without removing a duplicate. Inlining them back into `_validate_uuid_list` keeps each message next to the branch that emits it.

utils/decorators.py
```diff
@@ -133,7 +142,7 @@ def _resolve_region_jwt(
     if available_regions:
         return None, (
             f'Multiple regions available in token: {", ".join(available_regions)}. '
-            f'Please specify a region parameter.'
+            f'{_SPECIFY_REGION_HINT}'
```
junho226: Suggestion: this f-string only interpolates a constant.

Implicit concatenation of the plain name reads the same and drops the formatting round-trip (same at line 173).

```suggestion
            _SPECIFY_REGION_HINT
```

junho226 (dismissed):
## Review summary

Three copies of a four-line retry sequence collapsing into one closure, with the defect they were hiding named rather than quietly fixed. The body is unusually precise about what changed, and the parts I could check independently all held up.

### Verified independently

I reproduced the claimed checks rather than taking them on faith, and re-derived the equivalences from `ed6c7f3`.

| Check | Result |
| --- | --- |
| `uv run --extra dev python -m pytest tests -q` | **951 passed** |
| `uv run --extra dev ruff check .` | All checks passed |
| `uv run --extra dev ruff format --check .` | 82 files already formatted |
| `0960d49` code-motion purity | **line multiset identical** (581 lines, blanks dropped) |

Parity points I checked against the merge base:

- **`backoff()` arithmetic is an exact complement.** The original guarded on `retry_count < self.max_retries` after incrementing; `backoff` returns `False` on `retry_count >= self.max_retries`. Tracing `max_retries=3` through both gives 3 requests and 2 sleeps either way, which is what `test_get_500_error` pins with `call_count == 3`.
- **Every `_ERR_*` constant maps byte-for-byte** to the literal it replaced — I diffed all seven against their original sites (`'HTTP Error'`, `'Max retries exceeded'`, `'MFA Required'`, `'Request Error'`, `'Request Exception'`, `'Timeout'`, `'Unexpected Error'`).
- **`endpoint.startswith(_CACHEABLE_ENDPOINTS)` is exactly `any(endpoint.startswith(p) for p in ...)`** — `str.startswith` with a tuple is defined as the OR over its elements, so the seven prefixes and their order are irrelevant to the result.
- **The UUID extraction preserves the reporting order.** `_UUID_LIST_FIELDS = ('server_ids', 'servers')` matches the original block order, both messages are byte-identical, and the `format_validation_error` argument positions are unchanged.
- **`frozenset` vs `list` for `k not in`** is equivalent for string keys, and removes a per-call rebuild.

### Critical Issues (0)

None.

### Warnings (1)

**1. `utils/http_client.py:270` — the added `status_code` is not inert; it changes the response contract downstream, and the body does not name that**

The body says the in-place 5xx response "also carries `status_code`, which the old tail response dropped," framed as a payload detail. It is not — two consumers read that key:

- `utils/common.py:292` lifts it into `error_kwargs`, so `unwrap_http_result` now emits `status_code` on the error it returns to the tool caller.
- `utils/recovery_hints.py:220` reads it, and `utils/decorators.py:407` runs `enrich_error_response` over every decorated tool result — a live path, not a theoretical one.

The `message` change compounds it: `'Failed after N attempts'` → `'Server error after N attempts'` is both the text `unwrap_http_result` surfaces and an input to `_detect_error_domain`, which keys on the substring `server`.

I ran the two payloads through `enrich_error_response` to see what actually differs:

```
--- OLD ---
  keys : ['error', 'message']
  hints: None
--- NEW ---
  keys : ['error', 'message', 'recovery_hints', 'status_code']
  hints: ['The server encountered an internal error. Try again in a moment.',
          'If the issue persists, check server health.']
```

So a 5xx exhaustion now returns two keys it never returned before, and MCP clients read that envelope. This looks like a straight improvement — the agent gets actionable guidance where it previously got a bare string — and no test regressed because the three that pin this path assert only on `error`. But this PR's own standard is that every delta is named ("Two deliberate fixes, one added field, otherwise unchanged"), and this is a third delta flowing from the added field. Worth a line in the body and a test pinning `recovery_hints` on the 5xx-exhaustion response, so the next person to touch `message` learns they moved the domain detection.

### Suggestions (3)

**1. `utils/http_client.py:25` — `_BACKOFF_MULTIPLIER = 2` has one call site and reads worse than the literal**

`retry_delay = min(retry_delay * 2, self.max_retry_delay)` is already self-documenting sitting next to `max_retry_delay`; the named constant adds a lookup 167 lines away to learn that doubling is doubling. Of everything `29700ae` hoisted, only `_SPECIFY_REGION_HINT` has two call sites — `_CACHEABLE_ENDPOINTS` and `_SENSITIVE_LOG_KEYS` earn their place by removing a per-call rebuild, and the `_ERR_*` set is defensible as a contract catalog pinned by tests. This one buys neither.

**2. `utils/decorators.py:33-34` — the two UUID messages are single-use after the extraction they were hoisted out of**

`8788207` already deduped them by creating `_validate_uuid_list`; that function is now their only reader, so `29700ae` hoisting them to module level moves the text ~210 lines from its sole use site without removing a duplicate. Inlining them back into `_validate_uuid_list` keeps the message next to the branch that emits it.

**3. `utils/decorators.py:145,173` — `f'{_SPECIFY_REGION_HINT}'` is an f-string that only interpolates a constant**

Implicit concatenation of the plain name reads the same and drops the formatting round-trip: `f'Multiple regions available: ...' + _SPECIFY_REGION_HINT`, or just `_SPECIFY_REGION_HINT` as the second string in the existing implicit concat.

### Highlights

- **The 5xx exhaustion defect is named, not silently fixed.** The missing arm let the path fall into a tail that logged `Unexpected fallback` while three tests asserted it as the normal outcome — a log that had been lying about the most common failure mode. Collapsing the duplication is what made it visible, and the body says so rather than folding it into "refactor".
- **The retry-log reordering is correctly attributed.** All three pre-refactor copies logged `retrying (3/3)` before checking whether a retry followed, so this is a pre-existing defect surfaced by consolidation, not introduced by it — and the body checks the rest of the codebase for the same shape rather than just fixing the one instance.
- **The code-motion commit is separated and independently verifiable.** Keeping `0960d49` free of any edit is what let me confirm it with a line-multiset comparison in one command; had it been mixed into `ba7bb63`, the reordering would have had to be read by eye.
- **Two literals were deliberately left inline, with reasons that are load-bearing.** The PyJWT decode-options dict really is mutated by `_merge_options` when `verify_signature` is false, so hoisting it would have been a latent bug — that is exactly the kind of thing a "name every constant" pass gets wrong, and it was caught.
- **The inaccurate comment introduced in `29700ae` was found and corrected in `24d23d2`** rather than carried forward. Claiming callers branch on the error values when none do would have been a trap for whoever tried to change them.

---

**Verdict**: APPROVED

- [x] Refactor is faithful — backoff arithmetic, constants, and `_is_cacheable` equivalence re-derived from the merge base
- [x] Tests pass (951), `ruff check` and `ruff format --check` clean
- [x] No security vulnerabilities; no new dependencies
- [x] Code motion verified pure by line-multiset comparison
- [ ] One unnamed downstream behavior change (`recovery_hints` on 5xx exhaustion) — worth a body line and a test, not a merge blocker

utils/http_client.py
jisung-02: Named in the body and pinned by a test in `4e40dcc`.

I reproduced your finding first — `_detect_error_domain` reads the old payload as `general` (no hints) and the new one as `server` (two hints), so the message change is doing exactly what you said.

The body now has a paragraph naming this as a third deliberate delta rather than a payload detail, walking through `utils/common.py:292` -> `utils/recovery_hints.py:220` -> `utils/decorators.py:407` and the `server` substring that `_detect_error_domain` keys on.

`tests/test_http_client.py:142` drives the real exhaustion path and pins three things: `status_code` is on the response, `_detect_error_domain` reads the message as the `server` domain, and `enrich_error_response` attaches those exact two hints. Asserting only that `recovery_hints` is non-empty would not have caught a message edit that moves the domain to a different non-empty hint set, which is the failure mode you were pointing at.

utils/http_client.py
jisung-02: Agreed that it bought nothing, but I did not inline it — `ba16afb` moves it to `self.backoff_multiplier` in `__init__` instead.

Your three criteria are the right filter and this one fails all of them, but it also had a shape problem the inline fix would have left in place: the other three retry knobs (`max_retries`, `retry_delay`, `max_retry_delay`) are instance attributes at `utils/http_client.py:43-45`, so the multiplier was the only one of the four living as a module-level private constant. Putting it on the instance is what the value already was — a per-client tuning knob — and it becomes overridable like the rest.

`2.0` rather than `2` only to match the float knobs it now sits beside; the product was already a float.

utils/decorators.py
jisung-02: Inlined back into `_validate_uuid_list` in `ba16afb`.

You are right that `8788207` already removed the duplicate, so hoisting the text in `29700ae` only moved it away from its sole reader. The item message exceeds the line length inline, so it is implicit concatenation across two lines; I verified both strings are byte-identical to the constants they replaced.

I extended this to `_UUID_LIST_FIELDS` on the same reasoning — one call site, ~290 lines from the definition. It is back inline at the loop, and the note about the tuple order being the reporting order moved down with it so that fact stays next to the code that depends on it. Say the word if you would rather keep that one hoisted.

utils/decorators.py
jisung-02: Done in `ba16afb` — both sites now use `+ _SPECIFY_REGION_HINT`.

I checked whether the same shape exists elsewhere. `tools/token_tools.py:52` and `:400` have it too (`f'{_JWT_REQUIRED_NOTE}'` as the tail of a description block), and I initially changed those as well, then reverted. Two reasons: that file is untouched by this branch, so it widens a utils-only diff into `tools/`; and unlike the two sites here — where the preceding fragment is a real f-string, so the block was already mixed — its block is otherwise uniform plain literals that a lone `+` would break up. The remaining `_JWT_REQUIRED_NOTE` uses in that file (`:130`, `:184`, `:253`, `:345`) are ordinary f-strings with real text around the interpolation, and `utils/oauth.py:618` interpolates a variable, so neither is this shape.

Happy to do `tools/token_tools.py` as a separate PR if you think it is worth it.

tests/test_http_client.py
```diff
@@ -138,6 +139,36 @@ async def test_get_500_error(self, mock_httpx_client):
...
+            endpoint='/api/error/',
+            token='test-token',
+        )
+
+        assert result['status_code'] == 500
+        assert (
+            _detect_error_domain(result['status_code'], result['message']) == 'server'
+        )
+
+        enriched = enrich_error_response(dict(result))
```
junho226: Warning: this does not call `enrich_error_response` the way the live path does, and the domain assertion above it is inert at 500.

**`tool_name` is missing.** The live path is always `enrich_error_response(result, tool_name=func_name)` (`utils/decorators.py:404`). `_detect_error_domain` checks `tool_name` *before* the `server` substring in the message, so of the 164 tools wrapped by `mcp_tool_handler`, 18 do not resolve to `server` on the live path — 11 `file` (`webftp_*`, `list_webftp_logs`) and 7 `command` (`execute_command`, `*_command_acl`).

**But at 500 the domain changes nothing.** `_HINT_REGISTRY` has exactly one key for that status:

```
registry keys for status 500: [(500, 'general')]
distinct hint sets across all domains at 500: 1
```

Every domain misses the domain-specific lookup and falls through to `(500, 'general')`. So the rationale in your reply — that asserting only non-emptiness "would not have caught a message edit that moves the domain to a different non-empty hint set" — does not hold here: there is no different hint set to move to, and the exact-list assertion below already covers that case. What the domain assertion actually buys is a fast, self-naming failure if `server` ever leaves the message. Worth keeping; just not coverage.

The part worth acting on is forward-looking: the moment a `(500, 'server')` or `(500, 'file')` entry is added to `_HINT_REGISTRY`, this test keeps passing while those 18 tools start receiving different hints, and it would be pinning a path no caller takes. Passing `tool_name=` — ideally parametrized over one `server`-domain and one `command`-domain tool — makes it track the live path.

Not a blocker; production behavior is correct.

tests/test_http_client.py
```diff
@@ -138,6 +139,36 @@ async def test_get_500_error(self, mock_httpx_client):
...
+    ):
+        """5xx exhaustion carries status_code and a 'server'-domain message.
+
+        Both feed enrich_error_response, which every decorated tool runs over
+        its result, so changing either moves the hints the caller receives.
+        """
+        mock_response = create_mock_response(status_code=500)
+        mock_httpx_client.request.return_value = mock_response
+
+        result = await http_client.get(
```
junho226, line 171: Suggestion: this test really sleeps for 3 seconds.

```
3.00s call  test_get_500_exhaustion_enriches_with_server_hints
3.00s call  test_get_500_error
```

The whole suite is ~19s, so this one test is ~16% of it, doubling a cost `test_get_500_error` already pays — so it is a pre-existing pattern rather than one this PR introduces. Since what is under test is the post-exhaustion payload and not the timing, `monkeypatch.setattr(http_client, 'retry_delay', 0)` removes it. Use `monkeypatch` specifically: `http_client` is a module-level singleton, so a direct assignment would leak into the rest of the suite.

Fine as a follow-up.

junho226 (dismissed):
## Re-review summary

My previous review was dismissed by the two new commits, so this is a delta re-review of `24d23d2..ba16afb` only — I re-read that review and your four replies rather than re-deriving the whole branch.

All four points are addressed, and each reply checked the claim before acting on it rather than taking it on faith. Two of them came back better than what I asked for.

### Re-verified

| Check | Result |
| --- | --- |
| `uv run --extra dev python -m pytest tests -q` | **952 passed** (951 + the new one) |
| `uv run --extra dev ruff check .` / `ruff format --check .` | clean / 82 files already formatted |
| Inlined UUID messages vs. the constants they replaced | **byte-identical** (both) |
| `f'{_SPECIFY_REGION_HINT}'` -> `+ _SPECIFY_REGION_HINT` | resulting string identical |
| `_BACKOFF_MULTIPLIER = 2` -> `self.backoff_multiplier = 2.0` | schedule identical: `[1.0, 2.0, 4.0, 8.0, 16.0, 30.0]` |
| Dangling references to the four removed constants | none |
| PR body | the third delta now has its own paragraph |

Inlining `_UUID_LIST_FIELDS` was beyond what I raised, and moving the "tuple order is the reporting order" note down to the loop keeps that fact next to the code that depends on it. Order is preserved. Keep it as is.

### Critical Issues (0)

None.

### Warnings (1)

**1. `tests/test_http_client.py:166` — the new test does not call `enrich_error_response` the way the live path does, and its domain assertion is inert at 500**

Two things compound here.

**`tool_name` is missing.** The live path is always `enrich_error_response(result, tool_name=func_name)` (`utils/decorators.py:404`), but the test calls `enrich_error_response(dict(result))`. `_detect_error_domain` checks `tool_name` *before* the `server` substring in the message, so of the 164 tools wrapped by `mcp_tool_handler`, **18 do not resolve to `server`** on the live path — 11 `file` (`webftp_*`, `list_webftp_logs`) and 7 `command` (`execute_command`, `*_command_acl`).

**But at 500 the domain does not change the hints at all.** `_HINT_REGISTRY` has exactly one key for that status:

```
registry keys for status 500: [(500, 'general')]
distinct hint sets across all domains at 500: 1
```

Every domain misses the domain-specific lookup and falls through to `(500, 'general')`. So the rationale in your reply — that asserting only non-emptiness "would not have caught a message edit that moves the domain to a different non-empty hint set" — does not hold at this status code: there is no different hint set to move to, and the test already pins the exact hint list, which covers that case on its own. What the domain assertion actually buys is a fast, self-naming failure if `server` ever leaves the message. That is worth having; it just is not coverage.

The forward hazard is the part worth acting on: the moment someone adds a `(500, 'server')` or `(500, 'file')` entry to `_HINT_REGISTRY`, **this test keeps passing while those 18 tools start receiving different hints** — it would then be pinning a path no caller takes. Passing `tool_name=` (ideally parametrized over one `server`-domain and one `command`-domain tool) makes the test track the live path instead of a configuration production never uses.

Production behavior is correct either way; this is about the test and the claim attached to it, not a merge blocker.

### Suggestions (1)

**1. `tests/test_http_client.py:154` — the new test really sleeps for 3 seconds**

```
3.00s call  test_get_500_exhaustion_enriches_with_server_hints
3.00s call  test_get_500_error
```

The whole suite is ~19s, so this one test is ~16% of it, doubling a cost `test_get_500_error` already pays. Since what is under test is the post-exhaustion payload rather than the timing, `monkeypatch.setattr(http_client, 'retry_delay', 0)` removes it — via `monkeypatch` specifically, because `http_client` is a module-level singleton and a direct assignment would leak into the rest of the suite. Pre-existing pattern, so entirely fine as a follow-up.

### Highlights

- **Every one of the four items was reproduced before it was acted on.** For the `status_code` warning you re-derived the `general` -> `server` domain flip yourself first, and only then wrote the body paragraph and the test — which is why the disagreement below is about a registry detail and not about whether the effect is real.
- **The backoff constant came back better than my suggestion.** I asked for the literal inline; `self.backoff_multiplier` is the right answer, and the reason is one I missed: the other three retry knobs are already instance attributes at `utils/http_client.py:43-45`, so the multiplier was the only one of the four living as a module-level private. Moving it to the instance makes it what it already was, a per-client tuning knob. `2.0` for float parity with its neighbours is a fair call and provably changes nothing.
- **Scope was actively held.** You found the same f-string shape at `tools/token_tools.py:52` and `:400`, changed it, then reverted with two reasons that both hold — the file is untouched by this branch, and its block is uniform plain literals a lone `+` would break up. Checking the other four `_JWT_REQUIRED_NOTE` sites and `utils/oauth.py:618` to confirm they are a different shape is the part that makes the revert a judgement rather than an omission.
- **The one place you did not simply comply is the one that needed a second look.** Not inlining the multiplier, and asking rather than assuming about `_UUID_LIST_FIELDS`, is the right handling of review feedback.

---

**Verdict**: APPROVED

- [x] All four items from the dismissed review resolved
- [x] Restored literals byte-identical, backoff schedule unchanged, no dangling references
- [x] 952 tests pass; `ruff check` and `ruff format --check` clean
- [x] PR body now names the third delta
- [ ] New test does not mirror the live `tool_name` argument and sleeps 3s — follow-up, not a blocker

tests/test_http_client.py
jisung-02: Fixed in `283c326`. You are right on both counts, including that my stated rationale did not hold.

I confirmed the registry first: `[k for k in _HINT_REGISTRY if k[0] == 500]` is exactly `[(500, 'general')]`, and `get_server` / `execute_command` / `webftp_upload_file` / `None` all come back with the same two hints while resolving to `server` / `command` / `file`. So there is no other hint set to move to at this status code, and the claim I attached to the assertion was wrong.

The test now passes `tool_name=` and is parametrized over the three domains, asserting both the domain each resolves to and that the hints still converge. That makes it track the live `with_error_handling` call shape rather than a configuration production never uses.

I checked the forward hazard is actually caught rather than assuming it: injecting `_HINT_REGISTRY[(500, 'file')]` and running this test alone gives `1 failed`. Without `tool_name` it would have kept passing.

tests/test_http_client.py
jisung-02: Done in `2456f32`, and it turned out to be four tests rather than one.

`test_get_500_error`, `test_network_timeout`, and `test_connection_error` exhaust the retries the same way, so the four together were 12s of an 18.5s suite. They now share a `no_retry_delay` fixture using `monkeypatch.setattr(http_client, 'retry_delay', 0)` — `monkeypatch` for the reason you gave, and a separate fixture rather than folding it into `mock_httpx_client` because that one is used by ~40 tests and only these four sleep.

Suite is 18.5s -> 6.4s.

One knock-on: the domain test is now parametrized. Sharing a single request across the three domains was only ever to avoid paying the sleep three times, so with the sleep gone that reason — and the docstring line stating it — no longer held. This is also the form you originally asked for, and it names the failing domain in the case id.

Not touched: `tests/test_command_tools.py`'s 2s and 1s cases. Those are command polling rather than HTTP retry, and that file is untouched by this branch.

junho226 (approved):
## Re-review (delta): `ba16afb..2456f32`

Both items from my dismissed review are addressed. I re-read that review and checked only what changed rather than re-deriving the branch.

### Re-verified

| Check | Result |
| --- | --- |
| `uv run --extra dev python -m pytest tests -q` | **954 passed** (952 + 2 from the parametrize) |
| `ruff check .` / `ruff format --check .` | clean / 82 files already formatted |
| Suite wall-clock | **19s → 9.01s**; no `http_client` test remains in the slowest-6 |
| `('get_server','server')`, `('execute_command','command')`, `('webftp_upload_file','file')` | all three re-derived against `_detect_error_domain` — hold |
| `(500, …)` registry keys | still exactly `(500, 'general')`, so the convergence claim in the docstring is accurate |
| monkeypatch mechanism | `self.retry_delay` (`utils/http_client.py:44`) → local copy at `:178`, and `0 * backoff_multiplier == 0`, so every sleep is 0 and `request()` never mutates the instance |
| Any test asserting real timing | none |

**Warning 1 resolved, and the forward hazard is now actually guarded.** The parametrize is the right shape: `execute_command` and `webftp_upload_file` both carry `server` in the message yet resolve to `command` / `file`, so the rows *demonstrate* `tool_name` precedence rather than just asserting it. And the moment someone adds a `(500, 'server')` entry to `_HINT_REGISTRY`, the `get_server` row's exact-hints assertion breaks — which is precisely the failure I wanted from it.

**Suggestion 1 resolved, beyond what I raised.** I flagged the one new test; the fixture was applied to `test_network_timeout` and `test_connection_error` as well. The docstring naming the singleton-leak reason is the part that will stop the next person from replacing it with a plain assignment.

### Critical Issues (0)

None.

### Warnings (0)

None.

### Suggestions (1)

**1. `tests/test_http_client.py:44` — this fixture already exists in the repo, three directories away, in a strictly better form**

`tests/integration/conftest.py:200` has `no_sleep`, which solves the same problem differently:

```python
@pytest.fixture(autouse=True)
def no_sleep():
    ...
    with patch('utils.http_client.asyncio.sleep', side_effect=fast_sleep) as mock_sleep:
        mock_sleep.recorded_delays = recorded_delays
```

Three differences that matter:

- **autouse** — the integration suite runs 16 tests in 0.03s with nothing to remember. `no_retry_delay` is opt-in, so the next retry-exhausting test added to this file silently pays the 3s again, and nothing fails to tell anyone.
- **It stubs the sleep instead of zeroing its input**, so `recorded_delays` keeps the backoff schedule assertable. Zeroing `retry_delay` makes `[1.0, 2.0, 4.0, …]` unobservable — no test needs it today, but it forecloses the one test that would pin the schedule you just moved onto `self.backoff_multiplier`.
- **No singleton mutation at all**, so the caveat the docstring exists to explain simply disappears.

Hoisting `no_sleep` into `tests/conftest.py` as autouse would cover both suites and let this fixture and its four call-site edits go away. Follow-up, not a blocker — what is here works and halved the suite.

### Highlights

- **The rename earns itself.** `..._enriches_across_tool_domains` says what the test covers; the old name described a domain that turned out to be one of three.
- **The docstring states the registry fact rather than the outcome** — that `500` has only a `general` entry, so the domains converge *here*. That is the sentence that tells the next reader why three parametrized rows assert identical hints, which would otherwise read as copy-paste.
- **The fixture was applied wider than asked.** Extending it to the timeout and connection-error tests is where most of the 10s came from; only ~6s of it was in the two tests I named.

---

**Verdict**: APPROVED

- [x] Both items from the dismissed review resolved
- [x] 954 tests pass; `ruff check` and `ruff format --check` clean
- [x] Domain resolution and registry claims re-derived independently
- [x] Suite wall-clock halved with no timing assertion weakened
- [ ] `no_retry_delay` duplicates the existing `no_sleep` fixture — follow-up, not a blocker

Issue #144, opened 2026-07-29
Migrate to the MCP SDK 2.x protocol foundation


jisung-02:
Part of #248.

## Problem statement

Alpacon MCP Server is pinned to Python MCP SDK 1.x because SDK 2.x removes `mcp.server.fastmcp`, renames `FastMCP` to `MCPServer`, moves transport settings, and changes protocol result models. The pin keeps CI green but prevents the server from adopting MCP protocol revision `2026-07-28`.

The migration also exposes an HTTP composition problem. Streamable HTTP currently monkeypatches an SDK runner so authentication failures can escape the JSON-RPC response. The authentication middleware buffers the entire response, which cannot support a long-lived `subscriptions/listen` stream. Transport defaults also carry production behavior such as stateless HTTP and JSON responses, so moving to SDK 2.x without explicit settings could regress redeploy handling or MFA reauthorization.

This foundation must make the SDK and protocol transition without also replacing the production OAuth topology. The broader authorization, Auth0 CIMD, MFA claim, MRTR, cache, and domain-event work remains tracked by #248.

## Solution

Upgrade the dependency to Python MCP SDK `2.2.x`, migrate the high-level server to `MCPServer`, and update all public protocol-model interactions to SDK 2.x conventions.

Build Streamable HTTP as an explicit ASGI application, wrap it with streaming-safe authentication middleware, and run the composed app with uvicorn. Preserve stateless HTTP, JSON responses, OAuth routes, the current proxy, and its two-stage MFA behavior explicitly.

Support protocol revision `2026-07-28` as the official contract. Keep the SDK's automatic 2025-era negotiation as best effort and retain SSE as a deprecated transport. Publish safe server discovery metadata and prove authenticated subscription connection and cancellation without adding Alpacon domain-event delivery.

## User stories

1. As a modern MCP client user, I want the server to negotiate protocol revision `2026-07-28`, so that I can connect with the current protocol.
2. As a modern MCP client user, I want `server/discover` to identify the server and package version, so that my client can understand the connection before calling tools.
3. As a user of a 2025-era MCP client, I want the SDK compatibility path to remain available, so that the dependency upgrade does not cause an avoidable immediate outage.
4. As a local user, I want stdio to keep working, so that desktop and command-line integrations remain available.
5. As a remote user, I want stateless Streamable HTTP to keep working across redeploys, so that stale session identifiers do not break requests.
[...]

PR #145, opened 2026-07-29
Pin mcp below 2.0 to unblock CI

jisung-02:
## Background

The `mcp` Python SDK released [2.0.0](https://github.com/modelcontextprotocol/python-sdk/releases/tag/v2.0.0) on 2026-07-28, which drops the `mcp.server.fastmcp` package. Inspecting the 2.0.0 wheel, the package was renamed to `mcp.server.mcpserver` (its `__init__.py` docstring reads "MCPServer - A more ergonomic interface for MCP servers") and `mcp/server/fastmcp/` is gone. The release notes put it as "`FastMCP` is now `MCPServer`" and state the decorator API itself is unchanged.

The same notes prescribe exactly this fix for projects that are not ready yet: "If your project is not ready to migrate, keep a `<2` upper bound on your requirement (for example `mcp>=1.28,<2`)."

`pyproject.toml` declared `mcp>=1.28.1` with no upper bound, and the Test workflow installs with `uv pip install -e .[dev]`, which ignores `uv.lock`. So CI resolved 2.0.0 on the next run and `server.py:9` failed:

```
server.py:9: in <module>
    from mcp.server.fastmcp import FastMCP
E   ModuleNotFoundError: No module named 'mcp.server.fastmcp'
...
Interrupted: 21 errors during collection
```

Every test module that reaches `server.py` failed at collection. Nothing in this repo changed; the break came from upstream.

## Changes

`pyproject.toml`: `mcp>=1.28.1` to `mcp>=1.28.1,<2`, with a one-line comment naming the rename so the bound is not removed by accident. `uv.lock` re-generated with `uv lock`. The floor is unchanged; `mcp>=1.28.1` is what `main` already declared, set by 4d459b7 ("fix(deps): resolve Dependabot security alerts").

`CLAUDE.md`: the Dependencies section listed `mcp>=1.9.4` under a heading that says the values come from `pyproject.toml`. It was two steps stale — it missed 4d459b7 raising the floor to 1.28.1, and it would now also miss the bound. Updated to `mcp>=1.28.1,<2`.

The lockfile diff is 26 lines, but only one of them is the specifier. The other 24 drop `marker = "python_full_version ..."` from the `dependencies` entries of the pydantic packages, and that is uv-version churn rather than an effect of the bound. Isolating it: editing an unrelated specifier (`httpx`, leaving `mcp` alone) produces the same 26 lines and the same 12 marker removals, while a `pyproject.toml` edit that leaves `requires-dist` untouched produces none — so any re-resolution triggers it, regardless of what changed. Re-locking this exact change with uv 0.11.21 gives a 2-line diff and no marker removals, against 26 lines on uv 0.11.32, so the committed lock was written in the older format. `mcp` itself stays at 1.28.1 in the lock and the pydantic versions are unchanged, so nothing about the resolution changes semantically. Happy to re-lock on 0.11.21 to keep the diff at the specifier line if reviewers prefer the smaller diff.

## Testing

In a clean venv built the way CI builds one:

- `uv pip install --dry-run -e ".[dev]"` resolves `mcp==1.29.0`, not 2.0.0
- The 1.29.0 wheel still ships `mcp/server/fastmcp/{__init__,exceptions,server}.py`
- Full suite: 952 passed, 2 failed

The 2 failures (`tests/test_basic.py::test_import_main`, `tests/test_simple_integration.py::TestMCPServerConfiguration::test_main_entry_point`) reproduce on `main` without this change and are local-only. The repo root ships a committed `__init__.py`, so pytest walks the package boundary up past the repo root and prepends the repo's parent directory to `sys.path`; on my machine the clone directory is named `main`, so `import main` resolves to that directory package instead of `main.py`. CI clones into `alpacon-mcp`, so the name collision does not occur there.

## Follow-up

This only holds the 1.x line, and the 1.x line is now in maintenance mode: per the release notes it lives on the [`v1.x` branch](https://github.com/modelcontextprotocol/python-sdk/tree/v1.x) and receives critical bug fixes and security patches only. So the pin buys time, not an indefinite stay.

Migrating to the 2.0 API (`server.py`, `utils/decorators.py`, resource/prompt registration, the http/sse entry points, and the ASGI auth middleware) is tracked in #144, which is also where the `<2` bound should be dropped. #144 additionally notes that CI installing from `pyproject.toml` rather than `uv.lock` is what let an upstream major release reach the test job at all.

Upstream references for that work: [What's new in v2](https://py.sdk.modelcontextprotocol.io/whats-new/) and the [migration guide](https://py.sdk.modelcontextprotocol.io/migration/), which lists the breaking changes with before-and-after code. Beyond the rename, the notes flag several things worth checking against this repo: `httpx` and `httpx-sse` are replaced by `httpx2>=2.5.0` (we import `httpx` directly in `utils/http_client.py` and elsewhere), `mcp/types.py` becomes the separate `mcp_types` package with `mcp.types` kept as a permanent alias, `pydantic` moves to `>=2.12.0`, `pydantic-settings` is dropped, the websocket transports and `mcp.shared.session` are removed, and Streamable HTTP servers reject request bodies over 4 MiB with HTTP 413.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -40,7 +40,8 @@ classifiers = [
     "Environment :: Console",
 ]
 dependencies = [
-    "mcp>=1.28.1",
+    # mcp 2.0 renamed mcp.server.fastmcp to mcp.server.mcpserver; pinned until migrated
+    "mcp>=1.28.1,<2",
     "httpx>=0.27.1",
     "PyJWT[crypto]>=2.10.1",
 ]
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -540,7 +540,7 @@ Workflow guides that teach an agent the Alpacon operating discipline (Work Sessi
 ## Dependencies
 
 **Runtime dependencies** (from `pyproject.toml`):
-- `mcp>=1.9.4`: Model Context Protocol framework
+- `mcp>=1.28.1,<2`: Model Context Protocol framework (upper bound: 2.0 renamed `mcp.server.fastmcp` to `mcp.server.mcpserver`; see #144)
 - `httpx>=0.27.1`: Async HTTP client
 - `PyJWT[crypto]>=2.10.1`: JWT token verification
 
--- a/uv.lock
+++ b/uv.lock
@@ -40,7 +40,7 @@ requires-dist = [
     { name = "flake8", marker = "extra == 'dev'", specifier = ">=6.0.0" },
     { name = "httpx", specifier = ">=0.27.1" },
     { name = "isort", marker = "extra == 'dev'", specifier = ">=5.12.0" },
-    { name = "mcp", specifier = ">=1.28.1" },
+    { name = "mcp", specifier = ">=1.28.1,<2" },
     { name = "mypy", marker = "extra == 'dev'", specifier = ">=1.0.0" },
     { name = "pyjwt", extras = ["crypto"], specifier = ">=2.10.1" },
     { name = "pytest", marker = "extra == 'dev'", specifier = ">=7.0.0" },
@@ -596,10 +596,10 @@ resolution-markers = [
     "python_full_version < '3.14' and sys_platform != 'win32'",
 ]
 dependencies = [
-    { name = "annotated-types", marker = "python_full_version < '3.14'" },
-    { name = "pydantic-core", version = "2.33.2", source = { registry = "https://pypi.org/simple" }, marker = "python_full_version < '3.14'" },
-    { name = "typing-extensions", version = "4.14.0", source = { registry = "https://pypi.org/simple" }, marker = "python_full_version < '3.14'" },
-    { name = "typing-inspection", version = "0.4.1", source = { registry = "https://pypi.org/simple" }, marker = "python_full_version < '3.14'" },
+    { name = "annotated-types" },
+    { name = "pydantic-core", version = "2.33.2", source = { registry = "https://pypi.org/simple" } },
+    { name = "typing-extensions", version = "4.14.0", source = { registry = "https://pypi.org/simple" } },
+    { name = "typing-inspection", version = "0.4.1", source = { registry = "https://pypi.org/simple" } },
 ]
 sdist = { url = "https://files.pythonhosted.org/packages/00/dd/4325abf92c39ba8623b5af936ddb36ffcfe0beae70405d456ab1fb2f5b8c/pydantic-2.11.7.tar.gz", hash = "sha256:d989c3c6cb79469287b1569f7447a17848c998458d49ebe294e975b9baf0f0db", size = 788350, upload-time = "2025-06-14T08:33:17.137Z" }
 wheels = [
@@ -615,10 +615,10 @@ resolution-markers = [
     "python_full_version >= '3.14' and sys_platform != 'win32'",
 ]
 dependencies = [
-    { name = "annotated-types", marker = "python_full_version >= '3.14'" },
-    { name = "pydantic-core", version = "2.46.4", source = { registry = "https://pypi.org/simple" }, marker = "python_full_version >= '3.14'" },
-    { name = "typing-extensions", version = "4.16.0", source = { registry = "https://pypi.org/simple" }, marker = "python_full_version >= '3.14'" },
-    { name = "typing-inspection", version = "0.4.2", source = { registry = "https://pypi.org/simple" }, marker = "python_full_version >= '3.14'" },
+    { name = "annotated-types" },
+    { name = "pydantic-core", version = "2.46.4", source = { registry = "https://pypi.org/simple" } },
+    { name = "typing-extensions", version = "4.16.0", source = { registry = "https://pypi.org/simple" } },
+    { name = "typing-inspection", version = "0.4.2", source = { registry = "https://pypi.org/simple" } },
 ]
 sdist = { url = "https://files.pythonhosted.org/packages/18/a5/b60d21ac674192f8ab0ba4e9fd860690f9b4a6e51ca5df118733b487d8d6/pydantic-2.13.4.tar.gz", hash = "sha256:c40756b57adaa8b1efeeced5c196f3f3b7c435f90e84ea7f443901bec8099ef6", size = 844775, upload-time = "2026-05-06T13:43:05.343Z" }
 wheels = [
@@ -634,7 +634,7 @@ resolution-markers = [
     "python_full_version < '3.14' and sys_platform != 'win32'",
 ]
 dependencies = [
-    { name = "typing-extensions", version = "4.14.0", source = { registry = "https://pypi.org/simple" }, marker = "python_full_version < '3.14'" },
+    { name = "typing-extensions", version = "4.14.0", source = { registry = "https://pypi.org/simple" } },
 ]
 sdist = { url = "https://files.pythonhosted.org/packages/ad/88/5f2260bdfae97aabf98f1778d43f69574390ad787afb646292a638c923d4/pydantic_core-2.33.2.tar.gz", hash = "sha256:7cb8bc3605c29176e1b105350d2e6474142d7c1bd1d9327c4a9bdb46bf827acc", size = 435195, upload-time = "2025-04-23T18:33:52.104Z" }
 wheels = [
@@ -680,7 +680,7 @@ resolution-markers = [
     "python_full_version >= '3.14' and sys_platform != 'win32'",
 ]
 dependencies = [
-    { name = "typing-extensions", version = "4.16.0", source = { registry = "https://pypi.org/simple" }, marker = "python_full_version >= '3.14'" },
+    { name = "typing-extensions", version = "4.16.0", source = { registry = "https://pypi.org/simple" } },
 ]
 sdist = { url = "https://files.pythonhosted.org/packages/9d/56/921726b776ace8d8f5db44c4ef961006580d91dc52b803c489fafd1aa249/pydantic_core-2.46.4.tar.gz", hash = "sha256:62f875393d7f270851f20523dd2e29f082bcc82292d66db2b64ea71f64b6e1c1", size = 471464, upload-time = "2026-05-06T13:37:06.98Z" }
 wheels = [
@@ -1068,7 +1068,7 @@ resolution-markers = [
     "python_full_version < '3.14' and sys_platform != 'win32'",
 ]
 dependencies = [
-    { name = "typing-extensions", version = "4.14.0", source = { registry = "https://pypi.org/simple" }, marker = "python_full_version < '3.14'" },
+    { name = "typing-extensions", version = "4.14.0", source = { registry = "https://pypi.org/simple" } },
 ]
 sdist = { url = "https://files.pythonhosted.org/packages/f8/b1/0c11f5058406b3af7609f121aaa6b609744687f1d158b3c3a5bf4cc94238/typing_inspection-0.4.1.tar.gz", hash = "sha256:6ae134cc0203c33377d43188d4064e9b357dba58cff3185f22924610e70a9d28", size = 75726, upload-time = "2025-05-21T18:55:23.885Z" }
 wheels = [
@@ -1084,7 +1084,7 @@ resolution-markers = [
     "python_full_version >= '3.14' and sys_platform != 'win32'",
 ]
 dependencies = [
-    { name = "typing-extensions", version = "4.16.0", source = { registry = "https://pypi.org/simple" }, marker = "python_full_version >= '3.14'" },
+    { name = "typing-extensions", version = "4.16.0", source = { registry = "https://pypi.org/simple" } },
 ]
 sdist = { url = "https://files.pythonhosted.org/packages/55/e3/70399cb7dd41c10ac53367ae42139cf4b1ca5f36bb3dc6c9d33acdb43655/typing_inspection-0.4.2.tar.gz", hash = "sha256:ba561c48a67c5958007083d386c3295464928b01faa735ab8547c5692e87f464", size = 75949, upload-time = "2025-10-01T02:14:41.687Z" }
 wheels = [


jisung-02:
Note that this pin only covers installs from source. The latest published alpacon-mcp on PyPI (0.7.0) still declares mcp>=1.9.4 with no upper bound, so uvx alpacon-mcp currently resolves mcp 2.0.0 and fails to start with the same ModuleNotFoundError: No module named 'mcp.server.fastmcp'. I'll cut a release once this PR is merged so end users pick up the bound.

pyproject.toml
```diff
@@ -40,7 +40,8 @@ classifiers = [
     "Environment :: Console",
 ]
 dependencies = [
-    "mcp>=1.28.1",
+    # mcp 2.0 renamed mcp.server.fastmcp to mcp.server.mcpserver; pinned until migrated
+    "mcp>=1.28.1,<2",
```
junho226, line 44: Warning: this line also raises the floor from `mcp>=1.9.4` to `mcp>=1.28.1`, which the PR body does not mention (it describes the base as already being `mcp>=1.28.1`).

The floor — not the `<2` bound — is what produces the lockfile churn the body attributes to "the pin's own effect". Measured with uv 0.11.21, after confirming the control case (re-locking `main` unchanged gives a 0-line diff):

| specifier | lock diff | mcp | pydantic |
|---|---|---|---|
| `mcp>=1.9.4` (main) | 0 | 1.23.0 | 2.11.7 |
| `mcp>=1.9.4,<2` | 4 | 1.23.0 | 2.11.7 |
| `mcp>=1.28.1,<2` (this PR) | 197 | 1.29.0 | 2.11.7 + 2.13.4 |

Against the committed lock, `mcp` moves 1.23.0 → 1.28.1 (not "stays at 1.28.1") and `pydantic` gains a second resolution at 2.13.4, with `pydantic-core` gaining 2.46.4, for Python >= 3.14 — which is also where the new `resolution-markers` block comes from.

Suggestion (only if the floor bump was incidental rather than intended):

```suggestion
    # mcp 2.0 renamed mcp.server.fastmcp to mcp.server.mcpserver; pinned until migrated
    "mcp>=1.9.4,<2",
```

If it is intended, keep it — just say so in the body, since it also changes what `uvx alpacon-mcp` users get once you cut the release.

junho226 (dismissed):
## Summary

Correct diagnosis and the remedy upstream prescribes. I reproduced both the break and the fix:

- `.github/workflows/test.yml:43` installs with `uv pip install -e .[dev]`, which ignores `uv.lock` — an unbounded specifier is exactly what let 2.0.0 reach CI.
- `server.py:9` imports `mcp.server.fastmcp`.
- In a clean venv, `uv pip install --dry-run -e ".[dev]"` on this branch resolves `mcp==1.29.0`. CI is unblocked.

## Critical Issues (0)

## Warnings (1)

### 1. The floor bump is undisclosed, and it — not the `<2` bound — is what churns the lockfile

The body says the base was `mcp>=1.28.1` with no upper bound. It is `mcp>=1.9.4` (`pyproject.toml:43` on `main`). This PR also raises the floor by 19 minor versions, which the body never mentions.

That matters because the body attributes the lock diff to "the pin's own effect". I measured it with uv 0.11.21. Control first: re-locking `main`'s unchanged `pyproject.toml` gives a 0-line diff, so my uv agrees with the committed lock and the comparison below is clean.

| specifier | lock diff (lines) | mcp | pydantic |
|---|---|---|---|
| `mcp>=1.9.4` (main, control) | 0 | 1.23.0 | 2.11.7 |
| `mcp>=1.9.4,<2` | **4** | 1.23.0 | 2.11.7 |
| `mcp>=1.28.1,<2` (this PR) | 197 | 1.29.0 | 2.11.7 + 2.13.4 |

The `<2` bound on its own costs 4 lines. Everything past that is the floor's doing.

Two further body claims do not hold against the committed lock. `mcp` does not "stay at 1.28.1" — it moves 1.23.0 → 1.28.1. And the pydantic versions are not unchanged — `pydantic` gains a second resolution at 2.13.4 and `pydantic-core` gains 2.46.4, both for Python >= 3.14, which is also where the new `resolution-markers` block comes from.

If the floor bump is deliberate, keep it and say so in the body — but note its blast radius is wider than "unblock CI": it also changes what `uvx alpacon-mcp` users get once you cut the release you mentioned in the comments. If it is not deliberate, `mcp>=1.9.4,<2` is the minimal change and leaves the lock essentially untouched.

## Suggestions (1)

### 1. The lock and CI resolve to different `mcp` versions

The committed lock pins 1.28.1; a CI install from `pyproject.toml` resolves 1.29.0, and `uv lock` on this branch today also moves to 1.29.0 (the 197-line diff above). So the lockfile does not describe what CI actually tests. That is the same root cause already routed to #144, so it need not be solved here — but re-running `uv lock` before merge would at least keep the next contributor from inheriting unrelated churn.

## Highlights

- Names the rename in a comment next to the bound, so the bound is not dropped by accident later.
- Identifies the real root cause — CI installing from `pyproject.toml` rather than `uv.lock` — and routes it to #144 instead of widening this PR.
- Reports the 2 local test failures honestly, with the exact `sys.path` mechanism and why CI does not hit it.
- The follow-up section lists the concrete 2.0 migration surface (httpx2, `mcp_types`, dropped `pydantic-settings`, removed websocket transports, 4 MiB body cap) rather than just linking the guide.
- The PyPI note in the comments — 0.7.0 still ships `mcp>=1.9.4`, so `uvx alpacon-mcp` is broken until a release goes out — is the right catch and the right sequencing.

---

**Verdict: APPROVED**

The fix is correct and CI should not stay red on a description issue. Please correct the body before merge — it is what the release note will be written from.

- [x] Tests included and passing (CI-style clean-venv resolution verified)
- [x] No security vulnerabilities
- [ ] Documentation updated if needed — PR body needs the corrections above

jisung-02:
Thanks for going to the trouble of reproducing this — the measurements are what made me re-run my own experiment properly, and you were right that the body's explanation of the lockfile diff was wrong. I've corrected it. The cause turned out to be something other than the floor, though, and I think we may have been measuring against different trees. Details below.

## The base specifier

On `origin/main`, `pyproject.toml:43` is `mcp>=1.28.1`. `mcp>=1.9.4` was there historically but was replaced by 4d459b7 ("fix(deps): resolve Dependabot security alerts"). So this PR doesn't move the floor — the whole `pyproject.toml` diff is the `,<2` bound plus the comment line:

```
$ git show origin/main:pyproject.toml | sed -n 43p
    "mcp>=1.28.1",
$ git diff origin/main...fix/pin-mcp-below-2 --stat
 pyproject.toml |  3 ++-
 uv.lock        | 26 +++++++++++++-------------
```

Against the committed lock I also get `mcp` at 1.28.1 on both sides, and `pydantic` already resolved twice on `main` (2.11.7 and 2.13.4, with `pydantic-core` 2.46.4) with `resolution-markers` present in both files, so those aren't new here either.

I suspect the stale `mcp>=1.9.4` you were working from is ours, not yours: `CLAUDE.md`'s Dependencies section still listed it under a heading that says the values come from `pyproject.toml`, and the published `alpacon-mcp` 0.7.0 on PyPI also still declares `mcp>=1.9.4` in `requires_dist`. Either one reads as the current base. I've fixed the `CLAUDE.md` line in f3bc0f4 so the next reader doesn't hit the same thing.

## What actually causes the lockfile churn

You were right that "the pin's own effect" was the wrong attribution — my control case (re-locking `main` unchanged gives a 0-line diff) doesn't establish what I claimed it did, because uv doesn't rewrite the file at all when no re-resolution is needed. Isolating it properly:

| change | lock diff | marker removals |
|---|---|---|
| `main` unchanged (control) | 0 | 0 |
| `mcp>=1.28.1,<2` (this PR) | 26 | 12 |
| `mcp>=1.9.4,<2` | 26 | 12 |
| `mcp>=1.9.4` (floor lowered) | 26 | 12 |
| `httpx` specifier only, `mcp` untouched | 26 | 12 |
| comment-only edit, `requires-dist` untouched | 0 | 0 |

An edit that doesn't involve `mcp` at all produces the identical diff, so neither the bound nor the floor is the cause — any re-resolution is. And the re-resolution output is uv-version dependent:

| uv | lock diff for `,<2` | marker removals |
|---|---|---|
| 0.11.21 | 2 | 0 |
| 0.11.32 | 26 | 12 |

So the committed lock was written in the older format, and the 24 extra lines in mine are 0.11.32 emitting markers differently. The body now says this instead. Happy to re-lock on 0.11.21 if you'd rather the diff stay at the specifier line — it's a one-line change either way, so your call.

## On the suggestion

The lock/CI divergence is real and I've confirmed it: `uv pip install --dry-run -e ".[dev]"` resolves `mcp==1.29.0` while the lock pins 1.28.1, and none of the three `uv` steps in `test.yml` consults the lock.

One nuance on the remedy: `uv lock` doesn't move to 1.29.0 — I get 1.28.1 on both 0.11.21 and 0.11.32, because `uv lock` preserves already-locked versions unless the constraint invalidates them, and 1.28.1 still satisfies `>=1.28.1,<2`. `uv lock --check` exits 0 on this branch. So re-locking before merge wouldn't close the gap; that needs the CI install to go through the lock, which is the #144 item.

pyproject.toml
jisung-02: Going to leave the floor as it is, if that's alright — but the underlying point about the body was well taken, and I've rewritten that paragraph.

`origin/main`'s `pyproject.toml:43` is already `mcp>=1.28.1`; the `mcp>=1.9.4` you're comparing against was replaced by 4d459b7 ("fix(deps): resolve Dependabot security alerts"). So applying the suggestion would lower the floor rather than restore it, and reverts that Dependabot bump. Keeping `mcp>=1.28.1,<2`.

On the churn: it isn't the floor either. An edit to an unrelated specifier (`httpx`, `mcp` untouched) reproduces the same 26 lines and 12 marker removals, while uv 0.11.21 produces 2 lines and no marker removals for this exact change against 26 on 0.11.32 — it's re-resolution output differing by uv version. Full measurements and the `mcp>=1.9.4` provenance (stale in `CLAUDE.md`, now fixed in f3bc0f4, and still live in PyPI 0.7.0's `requires_dist`) are in https://github.com/alpacax/alpacon-mcp/pull/145#issuecomment-5112289189.

junho226: You are right and my Warning was wrong — retracting it, and the suggestion with it.

`origin/main:pyproject.toml:43` is `mcp>=1.28.1`, replaced by 4d459b7 (2026-07-16) which is an ancestor of main:

```
$ git merge-base --is-ancestor 4d459b7 origin/main && echo "in main: yes"
in main: yes
$ git show 4d459b7 -- pyproject.toml | grep -E "^[-+].*mcp"
-    "mcp>=1.9.4",
+    "mcp>=1.28.1",
```

So this PR does not move the floor, and everything I derived from that premise goes too — the 197-line lock diff, "mcp moves to 1.29.0", and the second pydantic resolution being new here. Your read on where my stale `1.9.4` came from is almost certainly right.

Your churn measurement reproduces on my side. From `main`'s lock with this PR's `pyproject.toml`, uv 0.11.21 writes only the specifier line (2 lines); the committed lock differs from that by 24, all redundant per-dependency markers your uv drops on entries whose `resolution-markers` already scope them.

On "happy to re-lock on 0.11.21 if you'd rather" — no need, and I checked the thing that would actually bite a contributor:

```
$ uv lock --check     # this branch, uv 0.11.21 -> exit 0
$ uv lock             # -> no-op, 0-line diff
```

0.11.21 accepts the committed lock and re-locking is a no-op, so both formats are fixpoints and there is no ping-pong. Leave it as committed.

junho226 (approved):
## Re-review (delta since f3bc0f4)

**My previous Warning 1 was wrong. Retracting it.** You are right on the base specifier, and I have verified the rest of your response independently.

### Retraction: there is no floor bump in this PR

`origin/main:pyproject.toml:43` is `mcp>=1.28.1`, exactly as you said:

```
$ git log -1 --format='%h %ci %s' 4d459b7
4d459b7 2026-07-16 20:30:35 +0000 fix(deps): resolve Dependabot security alerts
$ git merge-base --is-ancestor 4d459b7 origin/main && echo "in main: yes"
in main: yes
$ git show 4d459b7 -- pyproject.toml | grep -E '^[-+].*mcp'
-    "mcp>=1.9.4",
+    "mcp>=1.28.1",
```

So the `pyproject.toml` diff is the `,<2` bound plus the comment, nothing more. Your diagnosis of where my stale `1.9.4` came from is almost certainly right — the two places that still advertised it as current were `CLAUDE.md` and PyPI 0.7.0's `requires_dist`. Everything I derived from that premise falls with it: the 197-line lock diff, "mcp moves to 1.29.0", and the second pydantic resolution being new here. **Suggestion 1 goes too** — it rested on the same wrong base.

### Your churn measurement reproduces

Starting from `main`'s lock with this PR's `pyproject.toml`, uv 0.11.21 (mine) writes only the specifier line:

```
$ git show origin/main:uv.lock > uv.lock && uv lock
Resolved 56 packages in 905ms
$ diff <(git show origin/main:uv.lock) uv.lock
43c43
<     { name = "mcp", specifier = ">=1.28.1" },
---
>     { name = "mcp", specifier = ">=1.28.1,<2" },
```

2 lines, and the committed lock differs from that by 24 — the redundant per-dependency markers your uv drops (`{ name = "annotated-types", marker = "python_full_version < '3.14'" }` → `{ name = "annotated-types" }`, on entries whose `resolution-markers` already scope them). Confirms uv-version output, not the specifier.

### On "happy to re-lock on 0.11.21 if you'd rather" — no need

I checked whether the newer-uv format is stable under the older uv, since that is what would actually bite a contributor. It is:

```
$ uv lock --check     # on this branch, uv 0.11.21
Resolved 56 packages in 44ms          # exit 0
$ uv lock && diff uv.lock.orig uv.lock
                                      # no output — no-op
```

0.11.21 accepts the committed lock and re-locking is a no-op, so there is no ping-pong for anyone still on the older uv. Both formats are fixpoints; leave it as committed.

### The `<2` bound is load-bearing, verified against the real 2.0.0

Since this text now lives in both `pyproject.toml` and `CLAUDE.md`, I checked it against the published wheel rather than the changelog:

```
mcp.server submodules: [..., 'lowlevel', 'mcpserver', 'models', ...]
mcp.server.fastmcp:    FAILS -> ModuleNotFoundError: No module named 'mcp.server.fastmcp'
mcp.server.mcpserver:  IMPORT OK
```

`2.0.0` is the current PyPI release, so an unbounded specifier breaks `server.py:9` on the next clean install. The rename is stated correctly in both places, and #144 ("Migrate to mcp SDK 2.0 (mcp.server.fastmcp -> mcp.server.mcpserver)") is the right pointer.

### Remaining

Nothing blocking. The lock/CI divergence stands as you described it — `test.yml` never consults the lock, so CI resolves `mcp==1.29.0` against a lock pinning 1.28.1 — and you are right that re-locking does not close it, since `uv lock` keeps 1.28.1 while it satisfies the constraint. That is #144's `test.yml` half, correctly left out of this PR.

CI is green on all six checks. Thanks for pushing back with measurements instead of just applying the suggestion — the base-specifier error was mine and would have shipped a wrong release note.

---

**Verdict: APPROVE**

Issue #151, opened 2026-08-04
[FEATURE] Restructure the OAuth proxy: named constants, staged handlers, package split


jisung-02:
## Description

`utils/oauth.py` is a 1,600-line module that mixes three layers: the crypto layer (secret derivation, state signing, device ids, sealed grants, the nonce cookie), the redirect_uri allowlist, and the five HTTP route handlers with their fallbacks. The handlers are 100 to 250 lines each, repeat the same strings and numbers (endpoint paths, error codes, grant types, TTLs, limits), and close over `register_oauth_routes`. Restructure the module in three steps, none of which changes wire behavior:

1. Named constants: extract the repeated literals into module-level constants and share one definition with `tests/test_oauth.py` instead of duplicating them; hoist the remaining function-scoped imports.
2. Staged handlers: hoist the closures in `register_oauth_routes` to module level and split each handler into named stage functions (parse, validate, build or exchange, deliver). A stage rejects a request by raising one `_OAuthRequestError`, which the handler turns into the same response the inline `return` produced before. Order the module by concern, tighten comments and docstrings, and rename the functions whose names misstate their role (a `_check_*` that returns a bool, an `_oauth_error` that builds a response).
3. Package split: turn `utils/oauth.py` into `utils/oauth/` with `__init__.py` (`register_oauth_routes` and the fallbacks), `_http.py`, `_sealing.py`, `_redirect_uris.py`, and one module per handler. `server.py` keeps importing `register_oauth_routes` from `utils.oauth`. Split `tests/test_oauth.py` along the same lines and move each mock patch target to the module that now owns the call.

Code quality only; no behavior change. Verification: the full pytest suite, ruff, and a dump of the registered routes, the metadata response, and every constant value compared against `main`.

The old branch `refactor/oauth-magic-literals` is 90+ commits behind `main`; its work is redone on `151-oauth-named-constants`.

PR #152, opened 2026-08-04
fix(oauth): sign and verify the OAuth state parameter

jisung-02:
## Summary

The OAuth `state` was unsigned base64 JSON, so calling `/oauth/callback` with a forged state could steer the authorization code to an arbitrary redirect_uri. The state is now an HMAC-SHA256 signed envelope with a 10-minute expiry, and the callback rejects any state that fails verification with 400.

## Background

The callback trusts the `redirect_uri` inside the state and redirects the authorization code to it. Previously it only decoded the state without verifying a signature, and a lenient path echoed the raw string back on decode failure, letting forged states through.

## Changes

- `utils/oauth.py`: signing key derivation (`_get_state_secret`) — falls back to deriving from `AUTH0_CLIENT_SECRET` via HMAC when `ALPACON_MCP_STATE_SECRET` is unset. The derived key is identical across replicas, so no shared storage is needed. An explicit `ALPACON_MCP_STATE_SECRET` must be hex decoding to at least 32 bytes, so it cannot be weaker than the derived key it replaces (review feedback).
- `_sign_state`/`_verify_state`: `encoded.signature` envelope, constant-time comparison, 10-minute `exp`. The three producers (standard, MFA stage 1, stage 2) all go through one `_build_state` helper.
- Callback: a state that fails verification gets 400 `invalid_request`. A callback without any state is still accepted (Auth0 error callbacks can arrive bare, and an absent state names no redirect target to steer). Missing config surfaces as the same JSON 500 the other handlers return.
- Authorize: the state-building branch is wrapped in the handler's `ValueError` guard, so a malformed `ALPACON_MCP_STATE_SECRET` returns the message rather than a bare 500. Signing is where an operator hits the bad key first, and only the callback carried the diagnostic before (review feedback).
- `docs/configuration.md`: documents `ALPACON_MCP_STATE_SECRET`, its hex and 32-byte requirement, and `openssl rand -hex 32` as the way to generate one.

Cleanup, limited to what this branch introduced:

- Function-scoped imports in the tests hoisted to module level.
- Only the literals this branch added are named (`_STATE_SECRET_ENV`, the `mfa`/`regular` stage tags, `HTTPStatus` for the new callback response). Naming the pre-existing literals is split out to #151.
- The test mock server stub and the forged-state builder are each defined once instead of being copy-pasted per test.
- The tests patch the environment through `_STATE_SECRET_ENV` rather than repeating the env var name (review feedback).

## Safety

- A leaked state cannot be forged, and the expiry limits the replay window to 10 minutes.
- The state is a public value that reaches URLs, browser history, Referer headers, and proxy logs, which makes the signing key an offline brute-force target with no rate limit in front of it. Review demonstrated a five-character key recovered from a single observed state in about two minutes of single-core Python, so the explicit key is bounded on format rather than on character count — a 32-character typed passphrase passes a length check while its keyspace does not. This does not guarantee entropy (`'00' * 32` still passes); it removes the typed-secret failure mode the demonstration exercised.
- Stage-2 replay parameters keep the existing allowlist validation (defense in depth).
- Backward compatibility: an unsigned state issued right before deployment gets a 400 right after, but the client restarts from authorize and recovers immediately via the SSO session.

## Testing

Full `pytest` suite: 1007 passed (75 oauth tests, including the new signing, forgery, expiry, missing-config, weak-explicit-key, and malformed-key-on-authorize cases). ruff and mypy clean.

### DIFF
--- a/tests/test_oauth.py
+++ b/tests/test_oauth.py
@@ -60,25 +60,38 @@ def _set_oauth_env():
         yield
 
 
+class MockMCPServer:
+    """Stands in for FastMCP, collecting whatever register_oauth_routes registers."""
+
+    def __init__(self):
+        self.routes = []
+
+    def custom_route(self, path, methods=None):
+        def decorator(func):
+            self.routes.append(Route(path, func, methods=methods))
+            return func
+
+        return decorator
+
+
 @pytest.fixture
 def oauth_app():
     """Create a minimal Starlette app with OAuth routes registered."""
-    routes = []
-
-    class MockMCPServer:
-        def custom_route(self, path, methods=None):
-            def decorator(func):
-                routes.append(Route(path, func, methods=methods))
-                return func
-
-            return decorator
-
     mock_server = MockMCPServer()
-
     register_oauth_routes(mock_server)
+    return TestClient(
+        Starlette(routes=mock_server.routes), raise_server_exceptions=False
+    )
 
-    app = Starlette(routes=routes)
-    return TestClient(app, raise_server_exceptions=False)
+
+def _forge_state_under_valid_signature():
+    """A state whose payload was swapped out after signing, keeping the signature."""
+    signed = _sign_state({'redirect_uri': TRUSTED_REDIRECT_URI})
+    _, _, signature = signed.rpartition('.')
+    forged = base64.urlsafe_b64encode(
+        json.dumps({'redirect_uri': EVIL_REDIRECT_URI, 'exp': FAR_FUTURE_EXP}).encode()
+    ).decode()
+    return f'{forged}.{signature}'
 
 
 def _mock_auth0_response(status_code=200, json_data=None):
@@ -110,10 +123,6 @@ class TestStateSecret:
         assert _get_state_secret() == _get_state_secret()
 
     def test_registration_logs_state_secret_source(self, caplog):
-        class MockMCPServer:
-            def custom_route(self, path, methods=None):
-                return lambda func: func
-
         with caplog.at_level('INFO'):
             register_oauth_routes(MockMCPServer())
 
@@ -134,14 +143,7 @@ class TestStateEnvelope:
         assert payload['exp'] <= int(time.time()) + _STATE_TTL_SECONDS
 
     def test_tampered_payload_is_rejected(self):
-        signed = _sign_state({'redirect_uri': TRUSTED_REDIRECT_URI, 'state': ''})
-        _, _, sig = signed.rpartition('.')
-        forged = base64.urlsafe_b64encode(
-            json.dumps(
-                {'redirect_uri': EVIL_REDIRECT_URI, 'exp': FAR_FUTURE_EXP}
-            ).encode()
-        ).decode()
-        assert _verify_state(f'{forged}.{sig}') is None
+        assert _verify_state(_forge_state_under_valid_signature()) is None
 
     def test_unsigned_state_is_rejected(self):
         unsigned = base64.urlsafe_b64encode(
@@ -538,16 +540,6 @@ class TestOAuthToken:
 
     def test_token_fails_without_client_secret(self):
         """Token endpoint should return 500 when AUTH0_CLIENT_SECRET is missing."""
-        routes = []
-
-        class MockMCPServer:
-            def custom_route(self, path, methods=None):
-                def decorator(func):
-                    routes.append(Route(path, func, methods=methods))
-                    return func
-
-                return decorator
-
         env_without_secret = {
             'AUTH0_DOMAIN': TEST_AUTH0_DOMAIN,
             'AUTH0_CLIENT_ID': TEST_CLIENT_ID,
@@ -561,7 +553,7 @@ class TestOAuthToken:
             # Unset AUTH0_CLIENT_SECRET if present
             with patch.dict('os.environ', {'AUTH0_CLIENT_SECRET': ''}, clear=False):
                 register_oauth_routes(mock_server)
-                app = Starlette(routes=routes)
+                app = Starlette(routes=mock_server.routes)
                 client = TestClient(app, raise_server_exceptions=False)
                 response = client.post(
                     '/oauth/token',
@@ -805,16 +797,9 @@ class TestOAuthCallback:
 
     def test_callback_rejects_tampered_state(self, oauth_app):
         """Swapping the payload under a valid signature must be rejected."""
-        signed = _sign_state({'redirect_uri': TRUSTED_REDIRECT_URI})
-        _, _, sig = signed.rpartition('.')
-        forged = base64.urlsafe_b64encode(
-            json.dumps(
-                {'redirect_uri': EVIL_REDIRECT_URI, 'exp': FAR_FUTURE_EXP}
-            ).encode()
-        ).decode()
         response = oauth_app.get(
             '/oauth/callback',
-            params={'code': 'auth-code', 'state': f'{forged}.{sig}'},
+            params={'code': 'auth-code', 'state': _forge_state_under_valid_signature()},
             follow_redirects=False,
         )
         assert response.status_code == 400
--- a/tests/test_oauth.py
+++ b/tests/test_oauth.py
@@ -7,11 +7,27 @@ error handling).
 """
 
 import base64
+import hashlib
+import hmac
 import json
+import time
 from unittest.mock import AsyncMock, MagicMock, patch
+from urllib.parse import parse_qs, urlparse
 
 import pytest
+from starlette.applications import Starlette
+from starlette.routing import Route
 from starlette.testclient import TestClient
+
+from utils.oauth import (
+    _STATE_SECRET_INFO,
+    _STATE_TTL_SECONDS,
+    _build_state,
+    _get_state_secret,
+    _sign_state,
+    _verify_state,
+    register_oauth_routes,
+)
 
 # Test configuration constants
 TEST_AUTH0_DOMAIN = 'test.us.auth0.com'
@@ -40,9 +56,6 @@ def _set_oauth_env():
 @pytest.fixture
 def oauth_app():
     """Create a minimal Starlette app with OAuth routes registered."""
-    from starlette.applications import Starlette
-    from starlette.routing import Route
-
     routes = []
 
     class MockMCPServer:
@@ -54,8 +67,6 @@ def oauth_app():
             return decorator
 
     mock_server = MockMCPServer()
-
-    from utils.oauth import register_oauth_routes
 
     register_oauth_routes(mock_server)
 
@@ -79,30 +90,19 @@ class TestStateSecret:
     """Tests for state signing key derivation."""
 
     def test_explicit_env_secret_wins(self):
-        from utils.oauth import _get_state_secret
-
         with patch.dict('os.environ', {'ALPACON_MCP_STATE_SECRET': 'explicit-key'}):
             assert _get_state_secret() == b'explicit-key'
 
     def test_derives_from_client_secret_when_env_unset(self):
-        import hashlib
-        import hmac
-
-        from utils.oauth import _STATE_SECRET_INFO, _get_state_secret
-
         expected = hmac.new(
             TEST_CLIENT_SECRET.encode(), _STATE_SECRET_INFO, hashlib.sha256
         ).digest()
         assert _get_state_secret() == expected
 
     def test_derived_secret_is_deterministic(self):
-        from utils.oauth import _get_state_secret
-
         assert _get_state_secret() == _get_state_secret()
 
     def test_registration_logs_state_secret_source(self, caplog):
-        from utils.oauth import register_oauth_routes
-
         class MockMCPServer:
             def custom_route(self, path, methods=None):
                 return lambda func: func
@@ -117,24 +117,16 @@ class TestStateEnvelope:
     """Tests for state signing and verification."""
 
     def test_round_trip_preserves_payload(self):
-        from utils.oauth import _sign_state, _verify_state
-
         signed = _sign_state({'redirect_uri': 'https://claude.ai/cb', 'state': 'xyz'})
         payload = _verify_state(signed)
         assert payload['redirect_uri'] == 'https://claude.ai/cb'
         assert payload['state'] == 'xyz'
 
     def test_sign_adds_expiry(self):
-        import time
-
-        from utils.oauth import _STATE_TTL_SECONDS, _sign_state, _verify_state
-
         payload = _verify_state(_sign_state({'state': 'xyz'}))
         assert payload['exp'] <= int(time.time()) + _STATE_TTL_SECONDS
 
     def test_tampered_payload_is_rejected(self):
-        from utils.oauth import _sign_state, _verify_state
-
         signed = _sign_state({'redirect_uri': 'https://claude.ai/cb', 'state': ''})
         _, _, sig = signed.rpartition('.')
         forged = base64.urlsafe_b64encode(
@@ -145,37 +137,27 @@ class TestStateEnvelope:
         assert _verify_state(f'{forged}.{sig}') is None
 
     def test_unsigned_state_is_rejected(self):
-        from utils.oauth import _verify_state
-
         unsigned = base64.urlsafe_b64encode(
             json.dumps({'redirect_uri': 'https://evil.com/cb'}).encode()
         ).decode()
         assert _verify_state(unsigned) is None
 
     def test_wrong_signature_is_rejected(self):
-        from utils.oauth import _sign_state, _verify_state
-
         encoded, _, _ = _sign_state({'state': 'xyz'}).rpartition('.')
         assert _verify_state(f'{encoded}.bm90LWEtc2ln') is None
 
     def test_expired_state_is_rejected(self):
-        from utils.oauth import _sign_state, _verify_state
-
         with patch('utils.oauth.time.time', return_value=1000):
             signed = _sign_state({'state': 'xyz'})
         with patch('utils.oauth.time.time', return_value=99999999):
             assert _verify_state(signed) is None
 
     def test_malformed_state_is_rejected(self):
-        from utils.oauth import _verify_state
-
         assert _verify_state('not-a-state') is None
         assert _verify_state('') is None
 
     def test_non_ascii_state_is_rejected(self):
         """hmac.compare_digest raises TypeError on non-ASCII str, so compare bytes."""
-        from utils.oauth import _verify_state
-
         assert _verify_state('payload.서명') is None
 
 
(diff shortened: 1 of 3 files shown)


tests/test_oauth.py
```diff
@@ -7,18 +7,41 @@
...
 
+from utils.oauth import (
+    _STATE_SECRET_INFO,
+    _STATE_TTL_SECONDS,
+    _build_state,
+    _get_state_secret,
+    _sign_state,
+    _verify_state,
+    register_oauth_routes,
+)
```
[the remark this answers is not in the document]
jisung-02: Good catch, fixed in e00abdf. The test now imports `_STATE_SECRET_ENV` from `utils.oauth` and patches the environment with it, so a rename of the env var can no longer leave the test passing against a stale name.

jisung-02:
Addressed the review comment in e00abdf. Full suite still green: 1004 passed, ruff and mypy clean. Ready for another look.

choongbum (dismissed):
state 서명 구현 확인했습니다. HMAC-SHA256 + 상수시간 비교, 디코딩 전 서명 검증, 위조/만료/비ASCII 케이스 테스트까지 잘 커버돼 있습니다. 기존의 디코드 실패 시 raw state 에코백 경로를 400으로 바꾼 것도 적절합니다.

버그바운티 제보(#26JUL134) 대응 4개 항목 중 2번(state 서명)에 해당하는 조치로 이해했습니다. 나머지 1번(PKCE 강제), 3번(redirect_uri full-URI 화이트리스트), 4번(/oauth/register per-client)은 별건으로 진행 부탁드립니다. 특히 1번은 기존 클라이언트 호환성 확인이 선행돼야 하는 건이라 일정만 편하실 때 공유해 주시면 좋겠습니다.

utils/oauth.py
```diff
@@ -123,6 +141,68 @@ def _get_oauth_config() -> dict[str, str]:
...
 
 
+def _get_state_secret() -> bytes:
+    """Falls back to deriving from AUTH0_CLIENT_SECRET so no extra secret needs
+    provisioning; the derived key is identical across replicas, so state
+    verifies without shared server-side storage.
+    """
+    explicit = os.getenv(_STATE_SECRET_ENV, '')
+    if explicit:
+        return explicit.encode()
```
junho226: Warning: 명시적 secret에 강도 요구가 없어서, 설정하는 행위가 기본값보다 **약하게** 만들 수 있습니다.

파생 경로는 HMAC-SHA256 출력이라 항상 32바이트인데, 이 경로는 운영자가 넣은 문자열을 그대로 HMAC 키로 씁니다. 그리고 state는 **공개 값**입니다 — URL, 브라우저 히스토리, Referer, 프록시 로그, Auth0 로그에 남습니다. 공격자에게 오프라인 오라클이 생긴다는 뜻입니다.

관측된 state 하나로 5자 키를 복구해봤습니다.

```
observed state: eyJyZWRpcmVjdF91cmkiOiJodHRwczovL2NsYXVkZS5haS9hcGkvbWNwL2F1...
recovered key: b'mcp01' after 21,996,712 guesses in 119.4s (pure Python, single core)
forged state verifies against _verify_state logic: True
forged redirect_uri: https://attacker.example/steal
```

순수 파이썬 단일 코어로 2분입니다. hashcat/GPU면 1초 안쪽입니다.

**과장하지 않겠습니다.** 키를 복구해도 open redirect가 바로 열리지는 않습니다 — `_is_allowed_redirect_url`이 `claude.ai` / `chatgpt.com` / `chat.openai.com` 정확 일치와 loopback만 통과시키니 위 `attacker.example`은 거기서 막힙니다. 키 복구가 실제로 주는 건 **이 PR이 세우려는 성질 자체의 무효화**입니다. state 내용이 다시 공격자가 고르는 값이 되고 — PR 본문 표현대로 *"letting forged states through"* 상태로 돌아가고 — 남는 방어는 allowlist 하나입니다. 임의 포트·경로가 되는 loopback이 그 안에 있고요.

그리고 이게 조용히 일어납니다. 문서는 이렇게만 말합니다:

> set the optional `ALPACON_MCP_STATE_SECRET` to use an explicit key instead

강도 언급이 없고, 테스트도 `'explicit-key'`(12자)를 씁니다. 명시 값을 넣는 쪽이 더 약할 수 있다는 걸 알려주는 곳이 없습니다.

고치는 비용은 작습니다. 콜백이 이미 `ValueError` → 500을 처리하니 기존 모양에 그대로 맞습니다:

```suggestion
    explicit = os.getenv(_STATE_SECRET_ENV, '')
    if explicit:
        # The state travels in URLs and proxy logs, so a short key is an offline
        # brute-force target; the derived fallback below is always 32 bytes.
        if len(explicit) < 32:
            raise ValueError(f'{_STATE_SECRET_ENV} must be at least 32 characters')
        return explicit.encode()
```

문서 쪽은 `openssl rand -hex 32` 정도를 같이 적어주면 충분합니다.

junho226 (commented):
## Review

The signing scheme itself is put together correctly. I checked each of the places
this kind of change usually goes wrong:

- **Signature verified before decoding** — `hmac.compare_digest` gates `json.loads`,
  so the parser stays outside the trust boundary.
- **Expiry checked after the signature**, not before — reversed, the expiry check
  would be meaningless.
- `compare_digest` is fed bytes, with the reason in a comment (`str` raises
  `TypeError` on non-ASCII).
- `rpartition('.')` is safe: base64url never emits `.`, and a state with no dot
  leaves `encoded` empty and is rejected.
- Key derivation is domain-separated (`_STATE_SECRET_INFO`).
- The callback's `_is_allowed_redirect_url` re-validation is retained, so the
  signature is not the only gate.

Verified locally: `ruff check`, `ruff format --check`, and `mypy` clean;
`pytest` **1004 passed** (72 oauth). CI green on 3.12 / 3.13 / 3.14.

### Warnings (1)

**`utils/oauth.py:151` — the explicit secret has no strength requirement, so
setting it can be weaker than the default.** See the inline comment; it includes a
measured key recovery from a single observed state.

### Highlights

- Verify → decode → expire, in that order. All three are easy to get backwards,
  and the docstring even explains why the signature covers the encoded payload
  rather than the raw JSON.
- Documenting why `compare_digest` gets bytes. That comment is what stops the next
  person from "tidying" it into a `.decode()`.
- **Keeping a state-less callback accepted** — Auth0 error callbacks arrive bare,
  and an absent state names no redirect target to steer. An unconditional 400 here
  would have broken the legitimate error path.
- Folding the three producers into one `_build_state`. That is the shape that
  prevents a fourth producer from shipping without a signature.
- Justifying stateless verification from the property that the derived key is
  identical across replicas, rather than reaching for shared storage.
- Stating the deployment discontinuity honestly: a state issued moments before the
  rollout gets a 400, and the client recovers through the SSO session.
- Scoping the literal-naming cleanup to what this branch introduced and splitting
  the pre-existing ones to #151.

---

**Verdict**: COMMENTED — the scheme is right and the tests are thorough. The one
warning is about the parameter the whole security property rests on, so I would
like an answer on it before this merges.

utils/oauth.py
jisung-02: Agreed on the warning, fixed in f34759a — with one deliberate difference from the
suggestion, which I want to be explicit about.

Your key recovery is what decided the shape of the fix. The recovered key was
`b'mcp01'`, a typed string. A minimum character count does not keep that shape
out: `correct-horse-battery-staple-xyz` is 32 characters and passes, while its
keyspace is nowhere near 32 bytes of entropy. So the check rejects on format
first — the value must be hex, and must decode to at least 32 bytes:

```python
explicit = os.getenv(_STATE_SECRET_ENV, '')
if explicit:
    # The state travels in URLs and proxy logs, so the key is an offline
    # brute-force target; hex-only input keeps a typed passphrase out.
    try:
        key = bytes.fromhex(explicit)
    except ValueError:
        raise ValueError(
            f'{_STATE_SECRET_ENV} must be hex; '
            f'generate one with openssl rand -hex {_STATE_SECRET_MIN_BYTES}'
        ) from None
    if len(key) < _STATE_SECRET_MIN_BYTES:
        raise ValueError(
            f'{_STATE_SECRET_ENV} must decode to at least '
            f'{_STATE_SECRET_MIN_BYTES} bytes'
        )
    return key
```

The 32 comes from your own framing: the derived key is an HMAC-SHA256 output, so
it is always 32 bytes, and the explicit path must not land below the path an
operator gets by doing nothing. The bound is `>=`, not `==`, so a longer key is
still accepted. The error message names `openssl rand -hex 32`, which is exactly
what the docs now tell operators to run, so the failure carries its own fix.

What this does not do, to be clear: it does not guarantee entropy. `'00' * 32`
still passes. The entropy of a single string is not measurable — it is a property
of the distribution the string came from — and the heuristics that approximate it
(Shannon score, compression ratio) reject legitimate random keys often enough
that operators end up disabling the check. What the format bound does remove is
the dominant real failure mode, a human typing a memorable secret, which is the
one your demonstration exercised.

Two other options I looked at and dropped. Stretching the value with scrypt or
argon2 multiplies attacker cost without adding entropy, but `_get_state_secret()`
runs on every sign and verify, so it needs a cache this PR should not be
introducing. Validating at startup instead of lazily would catch a bad key
earlier, but it would mean reading the OAuth config at import time, which the
module deliberately does not do today.

One known cost: `openssl rand -base64 32` is now rejected even though it is not
weaker. The env var is new in this branch, so nobody is currently setting it, and
the error names the hex command directly. Noted in the commit trailers that
widening the parser is the right move if an operator hits it, rather than
dropping the format check.

Docs updated in the same commit: `docs/configuration.md` now states the hex
requirement, the 32-byte minimum, and the generation command, and says the value
is rejected at the first sign or verify rather than at startup.

Verified locally: `ruff check`, `ruff format --check`, `mypy utils/oauth.py`
clean; `pytest` 1006 passed, 74 in the oauth module — the two new ones cover the
non-hex and short-key rejections.

choongbum (dismissed):

utils/oauth.py
```diff
@@ -269,33 +367,22 @@ async def oauth_authorize(request):
...
-                {
-                    'redirect_uri': client_redirect_uri,
-                    'state': original_state,
-                }
-            )
-            composite_state = base64.urlsafe_b64encode(state_data.encode()).decode()
-
             params['redirect_uri'] = f'{server_url}/oauth/callback'
-            params['state'] = composite_state
+            params['state'] = _build_state(client_redirect_uri, original_state)
```
junho226: Warning: the new `ValueError` is only handled on the *verify* path, so a misconfigured key returns a bare 500 on the endpoint an operator hits **first**.

`oauth_authorize`'s `try/except ValueError` wraps only `_get_oauth_config()` (line 287); these `_build_state` calls at 370/385 sit outside it. Before this PR that was safe — `_get_state_secret` could only raise through `_get_oauth_config`, which the handler already called successfully above. The explicit-key check makes it newly reachable.

Reproduced with `ALPACON_MCP_STATE_SECRET='correct-horse-battery-staple'`:

```
AUTHORIZE (sign path)   -> 500  body: 'Internal Server Error'
CALLBACK  (verify path) -> 500  body: '{"error":"ALPACON_MCP_STATE_SECRET must be hex; generate one with openssl rand -hex 32"}'
```

The diagnostic you wrote never reaches the caller on the sign path; it only lands in the traceback. It fails closed, so this is not a security issue — but the commit body states the error "surfaces at the first sign or verify", and at first *sign* it surfaces without the message.

utils/oauth.py
```diff
@@ -781,5 +869,9 @@ async def oauth_register_fallback(request):
         return await oauth_register(request)
 
     logger.info(
```
junho226, line 883: Suggestion: this line already reads `os.getenv(_STATE_SECRET_ENV)` for the log, so validating at startup is nearly free here — one `_get_state_secret()` call turns the lazy 500 into a boot-time failure carrying the message, and makes the "at least 32 bytes" branch reachable in a way an operator can act on.

If you would rather keep it lazy per the commit's constraint, wrapping the two `_build_state` calls in `oauth_authorize`'s existing `except ValueError` gets the message into the response instead. Either way it is worth a test asserting the authorize endpoint returns the structured error, since that is the path an operator hits first.

junho226 (dismissed):
## Re-review of `f34759a`

`f34759a` answers the one warning I left: the explicit `ALPACON_MCP_STATE_SECRET` path now
parses the value as hex and requires at least 32 decoded bytes, matching what the derived
HMAC-SHA256 key always is. The 5-character key recovery I demonstrated is no longer
expressible — a typed passphrase is rejected on **format**, before length is even consulted,
which closes the exact shape the attack used.

The rejected alternatives in the commit body are the right calls. A character-count minimum
would pass the 32-character passphrase that was cracked, and entropy heuristics are not
measurable on a single string.

Verified locally at `f34759a`: `ruff check`, `ruff format --check`, `mypy` clean; `pytest`
**1006 passed** (74 oauth, +2 from the new cases). CI green on 3.12 / 3.13 / 3.14, CodeQL pass.

### Warnings (1)

**`utils/oauth.py:385` — the new `ValueError` is only handled on the verify path, so a
misconfigured key returns a bare 500 on the endpoint an operator hits first.** See the
inline comment; it includes the reproduction.

### Suggestions (1)

**`utils/oauth.py:871` — `register_oauth_routes` already reads the env var for its log line,
so a startup validation is nearly free there.** Inline.

### Highlights

- **Hex-only rather than a length rule.** Rejecting on format is what makes the check
  un-gameable; a length minimum would have passed the exact string I cracked.
- The error names `openssl rand -hex 32` and the documentation now tells operators to run
  precisely that. Both read `_STATE_SECRET_MIN_BYTES`, so the fix and the instruction cannot
  drift apart.
- `raise ... from None` suppresses the `bytes.fromhex` chain, so a malformed value cannot
  surface the operator's input through a `__cause__` traceback.
- Tests pin all three branches — valid hex, non-hex, one byte short — and
  `test_explicit_env_secret_wins` was rewritten to hex rather than left asserting the old
  passthrough behaviour.
- The env var is new in this PR (absent on `origin/main`), so the stricter parser cannot
  break an existing deployment. That is what makes hex-only affordable now rather than a
  migration.

---

**Verdict**: APPROVED — the warning I blocked on is resolved, and resolved correctly. What
remains is diagnostics, not the security property.

utils/oauth.py
jisung-02: Confirmed and fixed in `d8c51b8`. Your reading of the scope is exactly right: the `try` at line 287 covers only `_get_oauth_config()`, and before this PR that was sufficient because `_get_state_secret` could not raise on its own.

I wrapped the whole state-building branch rather than the two `_build_state` calls individually:

```python
        # _build_state signs with the configured key, so a malformed
        # ALPACON_MCP_STATE_SECRET raises here — on the endpoint an operator
        # reaches before the callback. Carry the message instead of a bare 500.
        try:
            if mfa_requested:
                ...
            else:
                ...
        except ValueError as e:
            return JSONResponse(
                {'error': str(e)}, status_code=HTTPStatus.INTERNAL_SERVER_ERROR
            )
```

Two `try` blocks around the calls alone would have duplicated the handler and would still miss a `ValueError` raised anywhere else in that branch later. The diff looks larger than the change is — most of it is the reindent, and the two reflowed lines (`stage2_authorize_params`'s tuple and the MFA `auth0_url`) are `ruff format` reacting to the new indentation, not edits.

This keeps validation lazy, per the `Constraint` in `f34759a`. It changes only what the caller sees, so the commit body's "surfaces at the first sign or verify" is now true on both paths rather than just one.

While in there: `JSONResponse` moved to the handler's top import line, replacing the three function-local copies of it, and the new handler uses `HTTPStatus.INTERNAL_SERVER_ERROR` rather than a bare `500`. The starlette imports stay function-local, matching every other handler in the module.

New test: `test_authorize_reports_malformed_state_secret` asserts the authorize endpoint returns the structured error — the test you asked for, on the path an operator hits first.

Verified locally: `ruff check`, `ruff format --check`, `mypy utils/oauth.py` clean; `pytest` 1007 passed, 75 in the oauth module.

utils/oauth.py
jisung-02: Taking the second of the two options you gave, and splitting the first out: `d8c51b8` wraps the state-building branch in `oauth_authorize`'s existing `ValueError` guard, with the test you asked for on the authorize path. Startup validation is filed as #156.

The split is deliberate rather than an attempt to do less. `f34759a` recorded "validation runs on the existing lazy path so it surfaces at first sign/verify rather than startup" as a `Constraint`, so moving the check into `register_oauth_routes` reverses a decision this PR made on the record. It also comes with a question that is not mechanical: with no explicit key, `_get_state_secret()` falls through to `_get_oauth_config()` and reads `AUTH0_CLIENT_SECRET`, so an unconditional call would break deployments that currently start without OAuth configured. Gating on the `os.getenv` result the log line already performs avoids that, but whether a malformed key should fail the boot or only warn is an operations policy call, not mine to make inside a security fix. #156 carries both questions and the regression test for the unset-key path.

You are right that it is nearly free there — the env read is already on that line. That is what makes it worth doing properly rather than folding it in here.

choongbum (approved):

junho226 (approved):
## Re-review of `d8c51b8`

The warning from my last pass is closed. Wrapping the whole state-building branch means a
malformed key now reaches the caller with the diagnostic instead of a bare 500. Re-ran the
same reproduction:

```
before:  AUTHORIZE -> 500  'Internal Server Error'
         CALLBACK  -> 500  '{"error":"ALPACON_MCP_STATE_SECRET must be hex; ..."}'

after:   AUTHORIZE -> 500  '{"error":"ALPACON_MCP_STATE_SECRET must be hex; generate one with openssl rand -hex 32"}'
         CALLBACK  -> 500  '{"error":"ALPACON_MCP_STATE_SECRET must be hex; generate one with openssl rand -hex 32"}'
```

**You picked the lazier of my two suggestions, and the reason you gave for rejecting the
other one is better than my suggestion was.** I listed startup validation in
`register_oauth_routes` first; the commit rejects it because `f34759a` recorded "validation
stays on the lazy per-request path" as a constraint, so validating at boot reverses a
decision this PR already made deliberately — and it would break startup for a deployment
that never touches the OAuth endpoints. Wrapping the branch rather than each `_build_state`
call is also the better of the two shapes: a future `ValueError` raised anywhere in that
branch is still carried.

Checked that no unbound-`auth0_url` path opens up: both branches assign it, and the `except`
returns immediately.

`ruff check`, `ruff format --check`, `mypy` clean; `pytest` **1007 passed** (+1). CI green on
3.12 / 3.13 / 3.14, CodeQL pass.

### Nits (1)

Inside the same handler, the `_get_oauth_config` branch still passes `status_code=500` while
the new one passes `HTTPStatus.INTERNAL_SERVER_ERROR` (`utils/oauth.py:288` vs `:397`). Same
value, two lines apart, two spellings. Not worth a push on its own — fold it in whenever you
next touch the file.

### Highlights

- **The test asserts the message, not just the status.** `assert 'must be hex' in
  response.json()['error']` is the actual regression point; a status-only assertion would
  have passed against the code before this commit.
- The test docstring says *why* it targets authorize ("signing is the first place a bad key
  raises"), which is what stops someone from later deciding a callback test covers it.
- Collapsing the three function-local `JSONResponse` imports into the handler's existing
  import line, while keeping starlette imports function-local as the rest of the module does.
  Fixing a review comment without rewriting the file's conventions around it.
- Recording both rejected options with their reasons. The startup-validation rejection is the
  part I got wrong in my own suggestion.

---

**Verdict**: APPROVED. One nit left, nothing blocking.

PR #157, opened 2026-08-05
fix(oauth): bind the OAuth state to the browser that started the flow

jisung-02:
## Background

The OAuth `state` is HMAC-signed and expires after 10 minutes (#152), so a callback's state is provably minted by this server and still fresh. That says nothing about which browser it was minted for. A state that arrives at `/oauth/callback` in a different browser than the one that started at `/oauth/authorize` is accepted just the same, and its code is forwarded to the client's redirect URI. The flow needs a value the callback can check against the browser actually in front of it.

## Changes

`/oauth/authorize` mints a per-flow nonce with `secrets.token_urlsafe(32)`, sets it in a `__Host-alpacon_oauth_nonce` cookie, and carries only its SHA-256 hash in the signed state as `nonce_hash`. `/oauth/callback` hashes the cookie it received and compares it with the hash inside the verified state. A mismatch, a state with no `nonce_hash`, or a request with no cookie is rejected with 400 `invalid_request`.

The check fails closed and sits before the `error` branch so the whole gate is in one place; an error callback carries no code, so nothing is lost by rejecting it there as well.

Cookie attributes and why each one is what it is:

| Attribute | Reason |
|---|---|
| `__Host-` prefix | the browser itself then requires `Secure` and `Path=/` and forbids `Domain`, so a sibling host cannot plant the cookie |
| `HttpOnly` | page scripts never read the nonce |
| `SameSite=Lax` | the return from Auth0 is a top-level navigation, and `Strict` would withhold the cookie on it |
| `Max-Age` = state TTL | cookie and state expire together |

Across the MFA round trip, stage 2 re-issues the state, so the cookie is re-set at the same point and the two expiries stay aligned. The cookie is cleared when the flow ends — on the redirect back to the client and on the JSON fallback — but deliberately not on a rejected callback, so a forced callback cannot wipe a live flow's cookie.

`_NONCE_COOKIE_ATTRS` holds the attributes shared by the set and delete helpers. The delete has to repeat `Path`, since a cookie is identified by name plus domain plus path, and `Secure`, since the browser rejects a `Set-Cookie` under a `__Host-` name without it and the cookie then survives the delete. The two helpers previously kept that match by comment rather than by code.

## Deployment

The `__Host-` prefix requires `Secure`, so `ALPACON_MCP_RESOURCE_URL` must be `https` and any proxy in front of the server must pass `Set-Cookie` and `Cookie` through. A deployment that does neither fails every callback with `invalid_request`. `docs/configuration.md` says so next to the state-signing bullet.

Known limit, not addressed here: the cookie name is fixed, so two flows started concurrently in the same browser overwrite each other's nonce and the earlier one fails at its callback. A per-flow cookie name would fix it.

## Testing

- `uv run pytest -q` — 1066 passed
- `uv run ruff check .` — All checks passed
- `tests/test_oauth.py`, 130 tests in the file: authorize sets the cookie with the four attributes, puts the matching hash in the state, and re-marks a browser that arrives with a stale cookie; callback accepts a matching cookie, rejects a mismatched one, rejects a browser with no cookie, and rejects a state carrying no `nonce_hash`; the cookie survives the MFA round trip and is re-set with the new state; it is cleared on both end-of-flow paths and left alone on a rejection
- The two guards that make `_nonce_cookie_matches` fail closed on a `nonce_hash` that is not a usable hash are pinned, both verified load-bearing by reverting them: without the `isinstance` check a forged non-string `nonce_hash` becomes a 500, and comparing `str` instead of `bytes` does the same for a non-ASCII one, since `hmac.compare_digest` raises `TypeError` on it
- The gate's placement ahead of the error branch is pinned too: an error callback carrying an unbound state comes back as `invalid_request` rather than being relayed to the client


utils/oauth.py
```diff
@@ -44,6 +46,20 @@
 # Covers an Auth0 login plus MFA; keeps a leaked state only briefly usable.
 _STATE_TTL_SECONDS = 600
 
+# The __Host- prefix makes the browser itself require Secure, Path=/ and no
+# Domain, so the cookie cannot be planted by a sibling host.
+_NONCE_COOKIE_NAME = '__Host-alpacon_oauth_nonce'
+
+# Shared by set and delete: a browser only drops a cookie when the attributes
+# match the ones it was set with. SameSite stays Lax because Strict drops the
+# cookie on the top-level return from Auth0.
```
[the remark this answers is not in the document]
jisung-02: Good catch, the comment was wrong about why. Reworded it in 6adc854.

Cookie identity is name + domain + path, so Path is the attribute the delete has to repeat to address the same cookie. Secure is a separate reason: under a `__Host-` name the browser rejects a Set-Cookie that omits it, so a delete without Secure never lands at all. HttpOnly and SameSite ride along without mattering for the delete, and the comment no longer claims otherwise.

utils/oauth.py
```diff
@@ -44,6 +46,21 @@
 # Covers an Auth0 login plus MFA; keeps a leaked state only briefly usable.
 _STATE_TTL_SECONDS = 600
 
+# The __Host- prefix makes the browser itself require Secure, Path=/ and no
+# Domain, so the cookie cannot be planted by a sibling host.
+_NONCE_COOKIE_NAME = '__Host-alpacon_oauth_nonce'
```
junho226, line 51: Warning: 본문의 "Known limit"을 재현했습니다. 쿠키 이름이 고정이라, 같은 브라우저에서 두 플로우가 겹치면 나중 것이 앞 것을 덮습니다.

```
flow1 authorize -> flow2 authorize (같은 브라우저)
flow1 cookie == flow2 cookie ?  False   (덮어씀)
flow1 callback                  ->  400 invalid_request
```

적어두신 대로 재시도로 복구됩니다. 사용자가 성급하게 두 번 눌러 생긴 경우라면 앞 플로우는 어차피 버리는 거라 무해하고요.

걸리는 건 **서로 다른 MCP 클라이언트가 같은 브라우저에서 동시에 붙는 경우**입니다. 한쪽이 이유 없이 실패하는데, 사용자가 받는 메시지가 `Invalid or expired state parameter`라 state를 의심하게 만듭니다. 실제로는 state가 멀쩡하고 쿠키가 덮인 건데요. (응답 메시지를 나누자는 얘기는 아닙니다. 어느 검사가 걸렸는지 공격자에게 알려줄 이유가 없으니 지금이 맞고, 로그가 `state not bound to this browser`로 구분해주니 디버깅 경로는 있습니다.)

막을 사안은 아닙니다. 다만 고치는 비용이 작아 보여서 — 쿠키 이름에 플로우별 접미사를 붙이고 그 접미사를 서명된 state에 같이 실으면, 콜백이 어느 쿠키를 볼지 알 수 있죠 — 이번에 닫을지 후속 이슈로 뺄지만 정해주시면 좋겠습니다.

후속으로 미룬다면 이슈에 위 재현을 그대로 넣어주세요. "동시 플로우가 깨진다"보다 "두 번째 authorize가 첫 번째의 쿠키를 덮어서 첫 번째 콜백이 400"이 다음 사람에게 훨씬 빠릅니다.

utils/oauth.py
```diff
@@ -937,7 +1007,11 @@ async def oauth_callback(request):
                 f'{config["auth0_base_url"]}/authorize?{urlencode(stage2_params)}'
             )
             logger.info('Stage 2: Redirecting to Auth0 regular audience (SSO)')
-            return RedirectResponse(url=auth0_url, status_code=302)
+            response = RedirectResponse(url=auth0_url, status_code=302)
+            # Stage 2 restarts the state expiry; re-set the cookie so the two
+            # do not drift apart.
+            _set_nonce_cookie(response, request.cookies[_NONCE_COOKIE_NAME])
```
junho226, line 1013: Nit: `.get()`이 아니라 직접 첨자인데, 지금은 안전합니다.

`stage`는 `state_data`에서만 오고 `state_data`는 게이트 뒤에서 읽히니, 게이트를 통과하지 못하면 stage 2에 도달할 수 없습니다. 그래서 여기 오는 요청은 `_nonce_cookie_matches`가 이미 True를 냈고, 그건 쿠키가 비어 있지 않다는 뜻이라 키가 반드시 있습니다. 확인했습니다.

다만 그 결합이 이 줄에서는 안 보입니다. 나중에 게이트를 옮기거나 stage 2 분기를 위로 끌어올리는 변경이 이 줄을 조용히 500으로 만들 수 있어요. `.get()`으로 바꾸면 빈 쿠키를 설정하게 되니 그건 답이 아니고, 한 줄 주석이면 충분합니다. 예를 들어 "게이트가 통과한 뒤라 쿠키는 반드시 존재한다" 정도요.

utils/oauth.py
```diff
@@ -345,6 +362,36 @@ def _build_state(redirect_uri: str, state: str, **extra) -> str:
...
+    return base64.urlsafe_b64encode(hashlib.sha256(nonce.encode()).digest()).decode()
+
+
+def _set_nonce_cookie(response: Response, nonce: str) -> None:
+    response.set_cookie(
+        _NONCE_COOKIE_NAME, nonce, max_age=_STATE_TTL_SECONDS, **_NONCE_COOKIE_ATTRS
+    )
+
+
+def _nonce_cookie_matches(request: Request, state_data: dict) -> bool:
```
junho226, line 381: 확인만 남깁니다 (수정 요청 아님). 이 함수의 fail-closed 분기를 하나씩 때려봤고, 유효하게 서명된 state 안에 `nonce_hash`를 별별 모양으로 넣어봐도 전부 400입니다.

```
no nonce_hash / None / int / list / "" / non-ASCII  ->  전부 400
```

`isinstance` 검사와 bytes 비교가 없으면 이 중 두 가지가 500이 된다고 본문에 적으셨는데, 되돌리지 않은 상태에서 400으로 떨어지는 것도 같이 확인됩니다. 서명이 유효한 state는 이 서버가 발급한 것이니 이 값들이 실제로 도달할 일은 없지만, 가드가 있는 이유는 "도달하지 않는다"에 기대지 않기 위해서죠.

`hmac.compare_digest`를 쓴 것도 맞습니다. 해시 비교라 타이밍이 실질적 위협은 아니지만, 이 파일의 다른 비교와 같은 원시함수를 쓰는 편이 낫습니다.

junho226 (approved):
## Review

This adds browser binding on top of the signed state. #152 proved a callback's state was minted
by this server and is still fresh; this adds that it was minted *for the browser in front of you*.
It is a security gate, so I spent the review on the fail-closed branches and ran all of them:

```
right browser                  -> 302  (works)
no cookie                      -> 400
other browser's cookie         -> 400
empty cookie                   -> 400
state with no nonce_hash       -> 400
nonce_hash = None / int / list -> 400   (not 500)
nonce_hash = ""                -> 400
nonce_hash = non-ASCII         -> 400   (not 500)
error callback + unbound state -> 400
```

**All eleven close.** The malformed-type cases answering 400 rather than 500 is the evidence that
the `isinstance` check and the bytes comparison are doing work — the PR body verifies them by
*reverting* them; this verifies them in place.

The cookie attributes, measured off the authorize response:

```
__Host-alpacon_oauth_nonce=…; HttpOnly; Max-Age=600; Path=/; SameSite=lax; Secure
```

All four are present and `Max-Age` equals `_STATE_TTL_SECONDS`.

`pytest` **1066 passed**; `ruff check`, `ruff format --check`, `mypy` clean; CI green on
3.12 / 3.13 / 3.14 with CodeQL passing.

### Warnings (1)

**`utils/oauth.py:51` — I reproduced the concurrent-flow limit the body discloses.** Inline, with
the reproduction. Not blocking; I would like to know whether it closes here or in a follow-up.

### Nits (2)

1. **`utils/oauth.py:1013`** — the stage 2 re-set subscripts `request.cookies` directly. Safe
   today, but the reason is not visible at that line. Inline.
2. **The deployment discontinuity is not in the body.** A state minted moments before this rolls
   out carries no `nonce_hash`, so its callback answers 400. #152 stated exactly this property for
   its own rollout ("a state issued moments before the rollout gets a 400, and the client recovers
   through the SSO session"); the same sentence applies here and is worth repeating, since this
   time the in-flight flow fails on a check the client cannot see.

### Highlights

- **The state carries only the hash; the nonce itself lives only in the cookie.** #152 established
  that the state travels through URLs and proxy logs, so putting the nonce there would have made
  the binding decorative. This carries that finding forward instead of relearning it.
- **The `__Host-` prefix.** It makes the browser itself enforce `Secure`, `Path=/` and no `Domain`,
  so "a sibling host cannot plant this cookie" is a browser rule rather than something this code
  has to keep getting right.
- **`SameSite=Lax` with the reason recorded.** `Strict` looks stronger and is wrong here — it
  withholds the cookie on the top-level return from Auth0 and breaks the whole flow. A case where
  the stricter value is the incorrect one needs its reason written down, or someone tightens it
  later.
- **The cookie is deliberately *not* cleared on a rejected callback.** Clearing there would let
  anyone wipe a live flow's cookie with one forced request — the cleanup path becoming the attack
  surface. That is an easy one to miss.
- **The gate sits ahead of the error branch.** An error callback carries no code, so nothing is
  lost by rejecting it there, and the gate stays in one place instead of two. Pinned by a test.
- **`_NONCE_COOKIE_ATTRS` shared between set and delete.** The delete must repeat `Path` (a cookie
  is identified by name + domain + path) and `Secure` (the browser rejects a `Set-Cookie` under a
  `__Host-` name without it, and the cookie then survives the delete). That match used to be held
  by a comment; now it is held by code.
- **The known limit is named by the author, with its fix.** The thing I reproduced was already in
  the body.
- The deployment requirements (`https`, and a proxy that passes `Set-Cookie` / `Cookie` through)
  are in `docs/configuration.md`. With `__Host-`, a deployment missing either fails *every*
  callback, so that note is load-bearing.

---

**Verdict**: APPROVED — the binding closes as intended and every fail-closed branch I could
construct is refused. The warning is something the body already concedes; I would just like the
disposition decided.

utils/oauth.py
jisung-02: Thanks for running the reproduction—agreed on the diagnosis, and we'll take this as a follow-up rather than closing it here. Filed as #160 with your reproduction included verbatim, along with the per-flow-suffix fix you sketched (suffix in the cookie name, carried inside the signed state so the callback knows which cookie to read). The issue also keeps your framing: "the second authorize overwrites the first flow's cookie, so the first callback gets 400".

utils/oauth.py
jisung-02: Agreed the invariant is invisible at this line. Deferring to #160 together with the other follow-up items—the fix there is the one-line comment you suggested ("the gate has passed, so the cookie must exist"), not a switch to .get(), for the reason you gave.

jisung-02:
@junho226 Thanks for the review—especially for re-running every fail-closed branch in place. Disposition on the three open items: all deferred to a follow-up, filed as #160.

1. **Concurrent-flow cookie overwrite (warning, `utils/oauth.py:51`)**: tracked in #160 with your reproduction included verbatim, framed as you suggested ("the second authorize overwrites the first flow's cookie, so the first callback gets 400"), together with the per-flow cookie-name-suffix fix.
2. **Stage-2 cookie access comment (nit, `utils/oauth.py:1013`)**: tracked in #160—a one-line comment stating the gate has already passed, not a switch to `.get()`.
3. **Rollout discontinuity**: tracked in #160—repeating the #152 sentence (a state minted moments before the rollout carries no `nonce_hash`, its callback answers 400, and the client recovers through the SSO session) in the PR body and deployment notes.

PR #163, opened 2026-08-06
docs: rewrite the documentation set against the shipped code

jisung-02:
## Background

The documentation had drifted from the code, in both directions. Tool listings named tools that no longer exist, omitted whole domains that do, described `eu1` as a served region, and passed parameters in shapes the input validator rejects. A reader following `docs/examples.md` would have written calls that fail before reaching the API.

The drift was structural, not incidental. `CLAUDE.md` was 838 lines and 41.4KB loaded on every session, and most of that volume restated what other files already said — a 275-line tool catalog duplicating `docs/api-reference.md`, a tool-adding recipe duplicating `CONTRIBUTING.md`, a dependency list duplicating `pyproject.toml`. `docs/installation-guide.md` and `docs/installation.md` covered the same task and had already reached conflicting instructions. Duplication at that size is drift waiting to happen.

Two smaller things surfaced while verifying claims against the source: the dev extra installed three linters nothing in the repo calls and omitted the one CI and the pre-commit hooks actually run, and the Task Master generator output committed in 8ab1aaa was unreferenced, with a `CLAUDE.md` import pointing into a gitignored directory.

## Changes

Every tool, parameter, resource URI, and environment variable in the guides was re-derived from the source rather than edited in place. Where a claim was wider than the implementation, it was narrowed to the actual behavior:

| Claim | What the code does |
|---|---|
| `get_top_servers(metric_types=[...])` | comma-separated string; endpoint keys are `disk_io` and `traffic`, not `disk`; always returns five |
| `page` / `page_size` on IAM list tools | integers, as the other eight paginated tools already documented |
| "every read tool has a resource" | six do not — `search_events`, `get_registration_guide`, `work_session_timeline`, `explain_approval_decision`, `webftp_check_status`, `health_check` — because a resource URI carries only path parameters |
| `uvx alpacon-mcp` starts the server | on a first run it runs the setup wizard and returns (`main.py:96-111`); the server starts on the next invocation |
| resource list | was missing `alpacon://server-notes/{region}/{workspace}/{note_id}`, which `tools/resources.py:75` registers |

`CLAUDE.md` is now 101 lines and 4.9KB. What stays is what reading the code will not tell you: the red lines, the development commands, six non-obvious invariants, and annotated pointers to the documents holding the detail. Two facts had no other home and moved to `docs/api-reference.md` rather than being dropped — why access control and security settings deliberately have no write tools, and which server endpoints do not exist. The MCP-only red line previously read "never use subprocess or shell commands" without qualification, which contradicted the same file's instruction to run `pytest`; it now scopes to how the agent reaches Alpacon.

`README.md` regained the reason host pinning exists. It had described pinning as protecting reachability, which is the smaller half: under ADR 0027 a workspace's URL slug is mutable and a freed slug can be reused by a different workspace, so re-deriving a host from a stale label can resolve to another tenant's host. The code comments never lost this (`utils/token_manager.py:189`, `utils/http_client.py:105`); only the user-facing text did. For a zero-trust tool the difference between "cannot connect" and "connects to the wrong workspace" is the whole point. `CLAUDE.md`'s pointer for this also went to `docs/configuration.md`, which carries only the `ALPACON_MCP_<REGION>_<WORKSPACE>_URL` line — it now points at README, where the object form of a token entry and the reason both live.

`docs/installation-guide.md` is deleted. Keeping one correct document beat reconciling two that had already diverged. `docs/examples.md` loses most of its volume for the same reason a wrong example costs more than a missing one: the bulk of it demonstrated tools that were never implemented.

Deleted as unreferenced: `.rules` and the four `.github/instructions/*.instructions.md` files, all output of the same Taskmaster generator, one of them with an `applyTo` pattern covering only the files being deleted. The `.gitignore` and `.dockerignore` entries for `.taskmaster` stay — removing them would let a local `.taskmaster/` be committed by accident, which is worse than two lines of residue.

`pyproject.toml` now installs `ruff` in the dev extra, at the lower bound `.pre-commit-config.yaml` already pins so the hook and a local run resolve to the same version, and drops `black`, `isort`, and `flake8`. Nothing in the repo calls those three; `[tool.ruff.lint]` already selects the rule sets they covered (`I` for import sorting, `F` and `E`/`W` for flake8's checks) with `[tool.ruff.format]` standing in for black.

## Safety

One code change rides along. `_NONCE_COOKIE_ATTRS` in `utils/oauth.py` is now a `TypedDict` instead of an inferred `dict[str, object]`, so unpacking it into `set_cookie` and `delete_cookie` type-checks. Nothing about the emitted cookie changes — same name, `Path`, `Secure`, `HttpOnly`, and `SameSite=Lax`. It is here because installing mypy in the dev extra made the local pre-commit hook start reporting 8 pre-existing `arg-type` errors that blocked every commit; CI had been carrying them all along because its mypy step is `continue-on-error: true`.

## Testing

- `uv run --no-sync pytest -q`: `1088 passed, 7 warnings in 6.13s`.
- `uv run --no-sync ruff check .`: `All checks passed!`
- `uv run --no-sync mypy --ignore-missing-imports --no-strict-optional .`: `Success: no issues found in 82 source files`. This is the first time the branch's mypy run is clean rather than reporting the 8 `utils/oauth.py` errors.
- `uv sync --extra dev` after the manifest change installs only ruff, mypy, mypy-extensions, and pathspec where it previously pulled black, isort, flake8, and their transitive deps.
- Claims in the rewritten docs were checked against the source, not against the previous text. Where a check could not be resolved from the code, the claim was removed rather than softened.

## Not fixed here

- `utils/error_handler.py:129-130,290` still accepts and advertises `eu1`, which Alpacon does not serve. The validator set, its error message, the tests, an example, and roughly 154 docstring lines all name it, so the cleanup is its own `fix(regions)` change rather than a docs edit.
- `ruff format --check .` reports `CONTRIBUTING.md` and `docs/troubleshooting.md`, because ruff reformats Python inside markdown fences and the installed version differs from the pinned v0.9.7. CI runs only `ruff check`, so nothing in the pipeline catches it; it predates this branch and is left alone.
- `pyproject.toml` carries two dev dependency lists — `[project.optional-dependencies] dev` and `[dependency-groups] dev` — with `pytest` in both at different lower bounds. Reconciling them changes which pytest a fresh checkout resolves, so it is left for a separate decision.


jisung-02:
Superseded: reopening with the scope narrowed to documentation files only. The manifest and `utils/oauth.py` changes that were in this branch have been dropped and are listed as follow-ups in the new PR.

PR #164, opened 2026-08-06
docs: rewrite the documentation set against the shipped code

jisung-02:
## Background

The documentation had drifted from the code, in both directions. Tool listings named tools that no longer exist, omitted whole domains that do, described `eu1` as a served region, and passed parameters in shapes the input validator rejects. A reader following `docs/examples.md` would have written calls that fail before reaching the API.

The drift was structural, not incidental. `CLAUDE.md` was 838 lines and 41.4KB loaded on every session, and most of that volume restated what other files already said — a 275-line tool catalog duplicating `docs/api-reference.md`, a tool-adding recipe duplicating `CONTRIBUTING.md`, a dependency list duplicating `pyproject.toml`. `docs/installation-guide.md` and `docs/installation.md` covered the same task and had already reached conflicting instructions. Duplication at that size is drift waiting to happen.

The Task Master generator output committed in 8ab1aaa was also unreferenced, with a `CLAUDE.md` import pointing into a gitignored directory that could never resolve anywhere but the machine that ran `task-master init`.

## Changes

Every tool, parameter, resource URI, and environment variable in the guides was re-derived from the source rather than edited in place. Where a claim was wider than the implementation, it was narrowed to the actual behavior:

| Claim | What the code does |
|---|---|
| `get_top_servers(metric_types=[...])` | comma-separated string; endpoint keys are `disk_io` and `traffic`, not `disk`; always returns five |
| `page` / `page_size` on IAM list tools | integers, as the other eight paginated tools already documented |
| "every read tool has a resource" | six do not — `search_events`, `get_registration_guide`, `work_session_timeline`, `explain_approval_decision`, `webftp_check_status`, `health_check` — because a resource URI carries only path parameters |
| `uvx alpacon-mcp` starts the server | on a first run it runs the setup wizard and returns (`main.py:96-111`); the server starts on the next invocation |
| resource list | was missing `alpacon://server-notes/{region}/{workspace}/{note_id}`, which `tools/resources.py:75` registers |

`CLAUDE.md` is now 101 lines and 4.9KB. What stays is what reading the code will not tell you: the red lines, the development commands, six non-obvious invariants, and annotated pointers to the documents holding the detail. Two facts had no other home and moved to `docs/api-reference.md` rather than being dropped — why access control and security settings deliberately have no write tools, and which server endpoints do not exist. The MCP-only red line previously read "never use subprocess or shell commands" without qualification, which contradicted the same file's instruction to run `pytest`; it now scopes to how the agent reaches Alpacon.

`README.md` regained the reason host pinning exists. It had described pinning as protecting reachability, which is the smaller half: under ADR 0027 a workspace's URL slug is mutable and a freed slug can be reused by a different workspace, so re-deriving a host from a stale label can resolve to another tenant's host. The code comments never lost this (`utils/token_manager.py:189`, `utils/http_client.py:105`); only the user-facing text did. For a zero-trust tool the difference between "cannot connect" and "connects to the wrong workspace" is the whole point. `CLAUDE.md`'s pointer for this also went to `docs/configuration.md`, which carries only the `ALPACON_MCP_<REGION>_<WORKSPACE>_URL` line — it now points at README, where the object form of a token entry and the reason both live.

`docs/installation-guide.md` is deleted. Keeping one correct document beat reconciling two that had already diverged. `docs/examples.md` loses most of its volume for the same reason a wrong example costs more than a missing one: the bulk of it demonstrated tools that were never implemented.

Deleted as unreferenced: `.rules` and the four `.github/instructions/*.instructions.md` files, all output of the same Taskmaster generator, one of them with an `applyTo` pattern covering only the files being deleted. The ignore entries that guarded that output go with it — `.gitignore`'s `.rules/` never matched anything, since the deleted `.rules` was a file and the pattern only matches a directory, and `.taskmaster` has no remaining reference in the repository once the generator output is gone.

## Scope

Documentation files only. No `.py` file, no manifest, and no lockfile is touched, so nothing here can change behavior — the diff is markdown, the extensionless `.rules`, and five ignore-file lines that referred to it. Everything the refresh surfaced in code or in packaging is listed under Follow-ups rather than fixed here.

Review feedback addressed, in the order it arrived:

- Three hard-coded values that duplicate what another file owns — the pinned `uvx alpacon-mcp@0.8.2` example and the 60% coverage floor quoted in two documents — now point at the releases page and `.github/workflows/test.yml` instead of restating a number that nothing keeps honest.
- `docs/api-reference.md` called `list_workspaces` the only tool that takes no `workspace`. `health_check` takes no parameters at all (`tools/health_tools.py:19`), so the claim was false; the line no longer counts tools.
- `docs/troubleshooting.md` showed `execute_command(command="ls -la /target/directory/")`. `server_id` and `workspace` are required (`tools/command_tools.py:284-287`), so copying it failed at the input validator. It now carries both, in the same call shape as `docs/examples.md`.

The last one is a class of defect, not one line, so the class was swept rather than the line patched — see Testing. That found one more: `work_session_create` takes `expires_at` and `description` as required parameters (`tools/work_session_tools.py:37-46`), while the reference labeled both optional and the `examples.md` flow omitted `expires_at`.

Then, from review:

- The rewrite replaced `black .`, `isort .`, and `flake8 .` with `ruff check .` and `ruff format .` in `docs/installation.md` and `CONTRIBUTING.md`. The three it removed are installed by the dev extra; `ruff` is not, so a working block became `command not found`. Both documents now install `ruff` and `pre-commit` with `uv tool install` before calling them. The manifest fix stays under Follow-ups.
- `.gitignore` and `.dockerignore` still carried entries for the generator output this branch deletes, and the `.rules/` pattern never matched the `.rules` file at all. Both are gone.
- `CLAUDE.md`'s region note said the validator's one further accepted value is the internal `dev`, while `utils/error_handler.py:130` is `{'ap1', 'us1', 'eu1', 'dev'}`. Naming `eu1` there would only be deleted by the follow-up, so the parenthetical stops enumerating validator contents and states what `dev` is instead — true whether or not #165 lands.
- The follow-ups below are filed as #165 through #170. A PR body stops being findable the day it merges; the `eu1` one especially needs to outlive this page, since 141 tool docstrings still read `Region (ap1, us1, eu1)`.

## Testing

- `git diff --name-only origin/main...HEAD` lists markdown files and `.rules` and nothing else, so the test suite, the linter, and the type-checker see exactly what they see on `main`. There is no behavior to test.
- Claims in the rewritten docs were checked against the source, not against the previous text: every tool name against its `@mcp.tool` registration, every parameter against the function signature and the input validator, every resource URI against `tools/resources.py`, every environment variable against the code that reads it. Where a claim could not be resolved from the code, it was removed rather than softened.
- After the review found a call example missing required arguments, that check was redone mechanically rather than by eye: the 58 tool-call examples across every markdown file and the per-tool parameter lists in `docs/api-reference.md` were parsed and compared against the signatures of all 166 registered tools, looking for a required parameter left out, a parameter named that no longer exists, and a required parameter labeled optional. It ends at zero mismatches. The candidates it raised and this branch rejected: group preambles that state shared parameters once ("Each takes `server_id`, `workspace`, and an optional `region`"), the `workspace` is always required convention at the top of the reference, brace-alternation shorthand in the resource URI list, and the numbered prose sketches in `docs/examples.md`, which abbreviate deliberately and already name every required argument.
- Removed docs were checked for orphaned links before deletion; `docs/installation-guide.md` had no inbound reference left once `docs/README.md` was rewritten.

## Follow-ups

Each of these was found while verifying the docs against the code, and each needs a code or manifest change, so none belong in a documentation-only PR.

- #165 — `utils/error_handler.py:129-130,290` accepts and advertises `eu1` as a region. `list_workspaces` returns only `ap1` and `us1` hosts, and the rest of the documentation names those two, so a caller passing `eu1` clears the validator and then fails at DNS on `{workspace}.eu1.alpacon.io`. Removing it touches 170 places across 24 files — the validator set, its error message, 141 tool docstrings, `config/token.json.example`, and 11 test sites that currently pin `eu1` as valid — and changes accepted input, so it is a `fix(regions)` change of its own. Whether Alpacon has an `eu1` region at all should be confirmed against the platform first; the evidence here is the repository's own docs plus the token's workspace list, not the platform's region catalog. The 141 docstrings matter more than the guides do: for an MCP client the tool description is the documentation, so until they change, the model still reads "Region (ap1, us1, eu1)" no matter what the guides say. Rejecting `eu1` at the validator is what closes this; documenting it here would only be deleted again by the same change.
- #166 — The dev extra never installed `ruff` or `pre-commit`, though both documents call them and CI runs `ruff` on every push, while `black`, `isort`, and `flake8` are installed and called by nothing. `[tool.ruff.lint]` already selects the rule sets those three covered. The documents work around it with `uv tool install`; moving the tools into the dev extra and dropping the three unused ones is a `pyproject.toml` and `uv.lock` change.
- #167 — `utils/oauth.py:58` infers `dict[str, object]` for `_NONCE_COOKIE_ATTRS`, which produces 8 `arg-type` errors when it is unpacked into `set_cookie` and `delete_cookie`. CI carries them because its mypy step is `continue-on-error: true`, but a local pre-commit hook running mypy blocks every commit. A `TypedDict` annotation fixes it without touching the emitted cookie.
- #168 — `ruff format` reaches into fenced Python inside markdown, but CI runs only `ruff check`, so nothing in the pipeline notices when a document drifts from it. `CONTRIBUTING.md` and `docs/troubleshooting.md` are brought in line here because the pre-commit hook rewrote them on every commit; any other document with Python blocks will do the same to whoever touches it next. Adding `ruff format --check` to CI would make that visible instead of surprising.
- #169 — The signature-versus-docs comparison that closed out the review feedback was run as a throwaway script and is not committed, so nothing stops the next parameter change from silently invalidating an example again. Landing it as a test or a CI step would, but it needs a decision about where it lives and how strictly it treats the prose sketches, which is a change to the pipeline rather than to a document.
- #170 — `pyproject.toml` carries two dev dependency lists — `[project.optional-dependencies] dev` and `[dependency-groups] dev` — with `pytest` in both at different lower bounds. Reconciling them changes which pytest a fresh checkout resolves, so it needs a decision rather than an edit.


docs/installation.md
```diff
@@ -34,13 +34,13 @@ Complete installation guide for the Alpacon MCP Server across different platform
...
 # Run Alpacon MCP Server directly
-uvx alpacon-mcp --help
+uvx alpacon-mcp
 
 # Run with environment variables
 ALPACON_MCP_AP1_PRODUCTION_TOKEN="your-token" uvx alpacon-mcp
 
 # Run specific version
-uvx alpacon-mcp@0.1.0
+uvx alpacon-mcp@0.8.2
```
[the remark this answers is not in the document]
jisung-02: Agreed, fixed in 3b16d8c. The example is now `uvx alpacon-mcp@<version>` with a comment pointing at the releases page, so there is no version number left in the document to go stale.

Worth noting the pin was accurate when written — v0.8.2 is the current release. That is the problem: a correct number nothing keeps correct is indistinguishable from a stale one until a reader hits it.

docs/installation.md
```diff
@@ -394,34 +387,38 @@ sudo chown -R alpacon-mcp:alpacon-mcp /opt/alpacon-mcp
...
 
-# Run linting
-black .
-isort .
-flake8 .
+# Type check
+mypy --ignore-missing-imports --no-strict-optional .
 ```
 
+CI runs the same three: `ruff check`, `mypy`, and `pytest` with coverage (the build fails below 60%).
```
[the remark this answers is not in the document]
jisung-02: Agreed, fixed in 3b16d8c. The line now states the invariant only — CI runs `ruff check`, `mypy`, and `pytest` under a coverage floor — and names `.github/workflows/test.yml` as where the current threshold lives.

The 60% was correct (`--cov-fail-under=60`, `test.yml:59`), which is exactly why it was worth removing: nothing would have failed when the workflow moved.

CONTRIBUTING.md
```diff
@@ -66,22 +65,20 @@ Thank you for your interest in contributing to the Alpacon MCP server! This guid
...
 We use the following tools for code quality:
 
-- **Black** for code formatting
-- **isort** for import sorting
-- **flake8** for linting
+- **ruff** for linting, import sorting, and formatting
 - **mypy** for type checking
 
-Run all checks:
+Both run as pre-commit hooks, and CI runs `ruff check`, `mypy`, and `pytest` with a 60% coverage floor.
```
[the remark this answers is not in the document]
jisung-02: Same fix in 3b16d8c, applied consistently with `docs/installation.md` so the two documents cannot disagree with each other either. Both now defer to `.github/workflows/test.yml` for the number.

jisung-02:
The suppressed comment is a real defect, fixed in 03ba8a0.

`health_check` takes no parameters at all (`tools/health_tools.py:19` — it is registered with a bare `@mcp.tool`, not through `@mcp_tool_handler`), so `list_workspaces` is not the only tool without a `workspace`. The line now says what a reader actually needs — why the parameter is absent — instead of counting which tools share the property:

> Get list of available workspaces. Takes no `workspace`—it is how you find the name every other tool needs.

Checked for the same claim elsewhere: `grep -rn "only tool" --include=*.md .` returned that one line and nothing else.

Branch is still documentation-only; `git diff --name-only origin/main...HEAD | grep -v "\.md$"` returns only `.rules`, which this PR deletes.

jisung-02:
Fixed in 2660dc1.

Confirmed: `execute_command` takes `server_id`, `command`, and `workspace` as required positional parameters (`tools/command_tools.py:284-287`), so the example as written failed at the input validator before any API call. `docs/troubleshooting.md:306` now reads:

```
execute_command(
    server_id="7e3984de-49ab-4cc6-bcdf-21fbd35858b8",
    command="ls -la /target/directory/",
    workspace="production",
)
```

Same call shape as `docs/examples.md:122`, so a reader moving between the two documents sees one form.

Checked the rest of the guides for the same defect rather than fixing the reported line alone: every other tool-call example in `docs/examples.md` (`get_cpu_usage`, `execute_command`, `execute_command_multi_server`, `webftp_upload_file`, `webftp_upload_content`, `webftp_download_file`, `create_alert_rule`) already passes `workspace` and its required ids.

The branch is still documentation-only—`git diff --name-only origin/main...HEAD` lists markdown plus the deleted `.rules` and nothing else.

docs/installation.md
```diff
@@ -394,34 +387,38 @@ sudo chown -R alpacon-mcp:alpacon-mcp /opt/alpacon-mcp
...
+# Install pre-commit hooks (ruff + mypy)
 pre-commit install
 
 # Run tests
-python -m pytest tests/
+pytest
+
+# Lint and format
+ruff check .
+ruff format .
```
junho226, line 410: **Suggestion:** this block swaps working commands for ones the documented setup does not provide.

After `uv pip install -e .[dev]`, it tells the reader to run `ruff check .` and `ruff format .`. `ruff` is in no dependency list:

```
[project.optional-dependencies] dev = pytest, pytest-asyncio, black, isort, flake8, mypy
[dependency-groups]           dev = pytest, pytest-asyncio, pytest-cov
uv.lock                           -> no ruff entry, no pre-commit entry
```

The previous text ran `black .`, `isort .`, and `flake8 .`, which the dev extra does install. So a block that worked when pasted now ends in `command not found`. `pre-commit install` on line 399 fails for the same reason — that line predates this PR, but it is the same defect.

This is the fact your Follow-ups already record ("the dev extra never installed `ruff`"). The `pyproject.toml` fix is out of scope here, agreed, but the gap can be closed inside the document: call it as `uvx ruff check .`, or put `uv tool install ruff pre-commit` above the block. `CONTRIBUTING.md:58` has the same bare `ruff check .`, though line 84 there offers `pre-commit run --all-files` as a way out.

By this PR's own standard — a wrong example costs more than a missing one — this is worth closing before merge.

CLAUDE.md
```diff
@@ -5,804 +5,99 @@ stack: python
...
+
+pytest                      # test suite
+python -c "from server import mcp; print('ok')"   # import smoke check
+```
+
+## Non-obvious invariants
+
+Things the code will not tell you at a glance:
+
+- **Regions are `ap1` and `us1` only** (the validator also accepts an internal
```
junho226: **Suggestion:** the validator's accepted set is missing `eu1`.

This reads "Regions are `ap1` and `us1` only (the validator also accepts an internal `dev`)", but the set is `{'ap1', 'us1', 'eu1', 'dev'}` (`utils/error_handler.py:130`).

The parenthetical exists to say what clears the validator beyond the served regions, and it omits the one value that actually produces a confusing failure: `eu1` passes validation and then dies at DNS, exactly the path your Follow-ups describe. Naming `eu1` alongside `dev` makes the line do the job it is here to do.

junho226 (dismissed):
## Summary

The whole documentation set was re-derived from the code rather than edited in place, and two overlapping documents were collapsed into one. `CLAUDE.md` goes 838 lines to 101. No Python file is touched.

## Independent verification

I re-ran the central claim of this PR mechanically against the source. Everything matches.

| Check | Result |
|---|---|
| Registered tools | 166 (`@mcp_tool_handler` + `@mcp.tool`) |
| Tool-call examples across all markdown | 58 sites, 0 missing required params, 0 unknown params |
| Tools absent from `api-reference.md` | 0 / 166 |
| Tools named in docs that do not exist | 0 (the 4 candidates are prompts in `tools/prompts.py`, and the docs label them as prompts) |
| Resource URIs | 76 registered, all covered via the `{a\|b}` shorthand list, `alpacon://server-notes/...` now present |
| "six read tools are tool-only" | computed READ_ONLY set minus resource-backed set == exactly the six named |
| Relative links and anchors | all resolve; zero references to any deleted file |
| `eu1` | 0 occurrences in documentation |

Individual facts check out against the source too: `get_top_servers` takes `metric_types: str = ''` with `cpu, memory, disk_io, traffic` (`metrics_tools.py:703`), IAM pagination is `int | None` (`iam_tools.py:40`), `work_session_create` requires `expires_at` and `description` (`work_session_tools.py:37`), the first-run wizard returns at `main.py:96-111`, logging goes to `sys.stderr` (`logger.py:31`), and neither `run_debug.py` nor a Websh tool exists. All four environment variables the docs name (`ALPACON_MCP_STATE_SECRET`, `ALPACON_MCP_REDIRECT_URI_REPORT_ONLY`, `ALPACON_MCP_TOOLSETS`, `ALPACON_WORK_SESSION`) are read by the code.

I also compared the headings of the deleted `docs/installation-guide.md` against the surviving `docs/installation.md`. Nothing is lost — the survivor is the wider of the two, covering Docker, systemd, and file permissions as well.

## Critical (0)

None.

## Suggestions (3)

Two inline, plus one on a file outside the diff. The first is worth taking before merge; neither blocks.

**`.gitignore:76` still carries `.rules/`.** The deleted `.rules` was a file and this pattern matches a directory, so it never covered it. The reason for keeping the `.taskmaster` entries — an accidental commit is worse than two lines of residue — does not apply here.

## Highlights

- Deriving claims from the source instead of from the previous text. Across 158 verification points in my own re-check, nothing disagreed. That is rare for a documentation PR.
- Sweeping the whole class of defect after review found one instance of it, rather than patching the one line. That is what surfaced the `work_session_create` required-parameter mislabel.
- Deleting one of two overlapping documents instead of reconciling them. Two documents that have already diverged cost more to keep than to merge.
- Restoring the host pinning rationale in `README.md`. The reason lived in `token_manager.py:189` and `http_client.py:105` all along; the user-facing text had lost it. For a zero-trust tool the distinction between "cannot connect" and "connects to another tenant" is the whole point, and the new text states it correctly.
- Listing the follow-ups without fixing them, especially the count behind `eu1` removal (170 sites across 24 files) and the note that the platform's region catalog should be confirmed first.

## On the follow-ups (not this PR)

For the `eu1` follow-up, the **141 tool docstrings** matter more than the documentation does. For an MCP client the tool description *is* the documentation, so removing `eu1` from the guides still leaves the model reading "Region (ap1, us1, eu1)". Right now the human-facing docs and the agent-facing docs disagree.

---

**Verdict: APPROVED**

- [x] Claims verified against source (independent mechanical re-check)
- [x] No behavior change (markdown plus `.rules` only)
- [x] No dangling references or broken links
- [x] Tool, parameter, resource, and env var inventories match the code

docs/installation.md
jisung-02: Agreed, fixed in ee3371b. You are right that the regression is this branch: `black`/`isort`/`flake8` were installed by the dev extra and the rewrite swapped them for a tool that is in no dependency list, so the block went from working to `command not found`.

Closed inside the document, as you suggested, in both places — `docs/installation.md` and `CONTRIBUTING.md` now run

```bash
uv tool install ruff
uv tool install pre-commit
```

before `pre-commit install` and the lint calls, which also fixes the `pre-commit install` line that predates this PR. Moving both into the dev extra and dropping the three unused formatters stays under Follow-ups, since it is a `pyproject.toml` and `uv.lock` change.

CLAUDE.md
jisung-02: Correct about the set — `utils/error_handler.py:130` is `{"ap1", "us1", "eu1", "dev"}`, and this line is one I wrote, so the omission is mine.

Naming `eu1` alongside `dev` was tried and then reverted, deliberately. Your own note on the follow-up is the reason: the 141 tool docstrings matter more here than the guides do, and while they still say `Region (ap1, us1, eu1)` a sentence in `CLAUDE.md` does not stop anything. The follow-up `fix(regions)` removes `eu1` from the validator, its error message, the docstrings, and the tests that pin it as valid — at which point this line is correct exactly as it stands, and a documented warning about `eu1` would only be deleted again by the same change.

So the fix is to make validation reject it, not to describe it, and that is out of scope for a documentation-only PR. Tracked in Follow-ups, which now carries your docstring point as the reason it leads.

jisung-02:
Thanks for the mechanical re-check — three items from the review are handled.

**`ruff` not installed** (ee3371b) — a regression this branch introduced, as you read it: the dev extra installs `black`, `isort`, and `flake8`, and the rewrite swapped the calls for `ruff` without adding the tool. `docs/installation.md` and `CONTRIBUTING.md` now `uv tool install ruff` and `uv tool install pre-commit` before using them, which also fixes the pre-existing `pre-commit install` line.

**`.gitignore:76`** (4112ccf) — taken, and extended. `.rules/` never matched the deleted `.rules` file, and `.taskmaster` has no remaining reference in the repository now that the generator output is gone, so the `.dockerignore` entry went too. The earlier reasoning for keeping them applied to a directory that could still be created; it does not apply to either of these.

**`eu1` in `CLAUDE.md`** — the set you name is right and the line is mine. Not documented, though: the follow-up removes `eu1` from the validator, which makes the line correct as written, so adding a sentence now means deleting it next PR. Your point about the 141 docstrings is why — the guides are not where this is decided. Full reasoning in the inline reply, and Follow-ups now leads with it.

The PR body records all three. Still documentation-only in effect: the diff is markdown, `.rules`, and five ignore-file lines.

CLAUDE.md
```diff
@@ -5,804 +5,99 @@ stack: python
...
+  `dev`). An omitted `region` is resolved from the JWT in remote mode and from
```
junho226: Suggestion: this is the one line on the branch that still contradicts the code, and the reasoning for leaving it covers the wrong half.

`utils/error_handler.py:130` is `{'ap1', 'us1', 'eu1', 'dev'}`. The lead clause is fine — `ap1` and `us1` really are the served regions. It is the parenthetical that makes a claim *about the validator*, and as such it is false today: the validator also accepts `eu1`.

I take the point from the PR body that naming `eu1` here would only be deleted by the follow-up. But that argument covers *adding* it, not the sentence as written — the sentence asserts something about the validator that is wrong whichever way the follow-up goes. Reworded so it stops enumerating validator contents, it costs nothing and survives either outcome:

```suggestion
- **Regions are `ap1` and `us1` only** (the internal `dev` is not a served
  region). An omitted `region` is resolved from the JWT in remote mode and from
```

docs/api-reference.md
```diff
@@ -585,20 +814,42 @@ Update workspace-wide preferences. Only the fields you provide are sent (partial
...
+
+There are likewise no user-settings or user-profile tools: `/api/user/settings/` and `/api/user/profile/` do not exist on the Alpacon server. The endpoints that do exist for adjacent data are `/api/profiles/preferences/`, `/api/workspaces/preferences/` (exposed above), and `/api/auth0/users/`. Role and permission management endpoints do not exist either, which is why the IAM section has no role or permission tools.
+
 ---
 
 
-## 🔍 Resources
+## 🔍 Resources and prompts
+
+Most read tools are also exposed as read-only MCP resources under the `alpacon://` scheme, so a client can pull data without a tool call. `search_events`, `get_registration_guide`, `work_session_timeline`, `explain_approval_decision`, `webftp_check_status`, and `health_check` are tool-only. The URI convention is `alpacon://<domain>[/<sub>]/{region}/{workspace}[/{id}]`; optional filters are not part of the URI, so resources use each tool's defaults.
```
junho226, line 832: Suggestion: file the six follow-ups as issues before this merges.

I checked the open issue list and none of them exist yet — nothing for the `eu1` validator, the missing `ruff`/`pre-commit` in the dev extra, the `oauth.py` `TypedDict`, `ruff format --check` in CI, the signature-comparison script, or the two dev dependency lists in `pyproject.toml`.

The `eu1` one carries the most weight, and this line is why. The guides here are now correct, but 141 tool docstrings still read "Region (ap1, us1, eu1)" — and for an MCP client the tool description *is* the documentation. Until the validator change lands, the guides say one thing and the surface the model actually reads says another, which is the same drift class this PR is closing.

A PR body stops being findable the day it merges. Six issues linked from here would keep them.

junho226 (dismissed):
## Summary

Docs re-derived from source rather than edited in place, plus deletion of unreferenced Taskmaster generator output. `CLAUDE.md` drops from 838 lines to 101. `docs/installation-guide.md` is deleted in favor of the one correct sibling.

## Verification I ran

I reproduced the mechanical check the PR body describes, independently — AST-parsing all `@mcp_tool_handler` / `@mcp.tool` registrations and comparing them against every tool-call example across all markdown:

```
registered tools: 166
tool-call examples parsed: 58
missing required params:   0
unknown params:            0
registered tools never named in any doc: 0
doc names written as calls but not registered: 5
  -> all 5 are real: 4 @mcp.prompt() in tools/prompts.py,
     validate_file_path at utils/error_handler.py:153
```

Spot-checks of the claim table, each against source:

| Claim | Verified at |
|---|---|
| `get_top_servers` takes a comma-separated string; keys `disk_io`/`traffic`; always five | `tools/metrics_tools.py:703-714` |
| six read tools are tool-only | computed independently: `explain_approval_decision`, `get_registration_guide`, `health_check`, `search_events`, `webftp_check_status`, `work_session_timeline` — exactly the six named at `docs/api-reference.md:832` |
| `uvx alpacon-mcp` runs the wizard and returns on first run | `main.py:99-111` — the `return` before server start |
| `health_check` takes no parameters | `tools/health_tools.py:19` |
| `work_session_create` requires `expires_at` and `description` | `tools/work_session_tools.py:37-46` |
| `alpacon://server-notes/{region}/{workspace}/{note_id}` was missing | `tools/resources.py:75` |
| host pinning is about wrong-tenant resolution, not reachability | `utils/token_manager.py:187-194`, `utils/http_client.py:102-107` — README now matches |

Resource coverage: expanding the docs' brace-alternation shorthand and the stated `[/{id}]` convention accounts for all 77 registered URIs. No broken relative links in any markdown. `uvx ruff format --check` reports 12 files already formatted. `git diff --name-only origin/main...HEAD` is markdown plus `.rules` and two ignore files — no `.py`, no manifest, no lockfile, so the claim that nothing here can change behavior holds. No remaining reference to `installation-guide`, `.taskmaster`, or `.rules` anywhere in the tree.

## Critical issues (0)

None.

## Warnings (0)

None.

## Suggestions (2)

**1. `CLAUDE.md:44` — the one line still contradicting the code.** `utils/error_handler.py:130` is `{'ap1', 'us1', 'eu1', 'dev'}`, so "the validator also accepts an internal `dev`" enumerates what else the validator takes while omitting `eu1`. The lead clause is fine — `ap1` and `us1` really are the served regions. It is the parenthetical, which makes a claim *about the validator*, that is false today. Rewording so the sentence stops enumerating validator contents costs nothing and survives the follow-up either way, rather than depending on it landing.

**2. File the six follow-ups as issues before this merges.** None of them exist yet — I checked the open list, and there is no issue for the `eu1` validator, the missing `ruff`/`pre-commit` in the dev extra, the `oauth.py` `TypedDict`, `ruff format --check` in CI, the signature-comparison script, or the two dev dependency lists. The `eu1` one carries the most weight: 141 tool docstrings still read "Region (ap1, us1, eu1)", and for an MCP client the tool description *is* the documentation. Until that lands, the guides say one thing and the surface the model actually reads says another — the same drift class this PR is closing. A PR body stops being findable the day it merges.

## What worked well

- Re-deriving every claim from source instead of editing prose is why the mechanical check comes back at zero. I ran it independently and got the same result, including the correctly-rejected candidates (group preambles, brace shorthand, positional prose sketches in `examples.md`)
- The judgment on `eu1` — that documenting it would only be deleted by the follow-up — is the right instinct, applied one line too far
- Deleting `installation-guide.md` rather than reconciling two documents that had already diverged. Duplication at that size is drift waiting to happen, and the PR says so
- Replacing hard-coded values (`0.8.2`, the 60% floor) with pointers to the releases page and `.github/workflows/test.yml:59`, where something actually keeps them honest
- Catching that the rewrite's own `ruff` swap broke a working command block, and fixing it with `uv tool install` rather than pretending the manifest was already right
- Follow-ups scoped by *what kind of change they need*, not by effort. The `eu1` entry even states what evidence it lacks — the platform's region catalog, versus the repository's own docs

## Checklist

- [x] Follows stack conventions
- [x] Tests (n/a — no `.py` changed; verified by diff)
- [x] No security issues
- [x] Documentation updated (this is the documentation)

CLAUDE.md
jisung-02: Taken as suggested, in 2e0f1df.

You are right that my reasoning covered the wrong half. "Adding `eu1` would only be deleted by the follow-up" is an argument against adding it, not an argument for the sentence as it stood — the parenthetical made a claim about the validator, and that claim is false today regardless of which way the follow-up goes. Worse, if the follow-up slips, the false line stays indefinitely, which is the drift class this PR exists to close.

The line now reads:

```
- **Regions are `ap1` and `us1` only** (the internal `dev` is not a served
  region).
```

It stops enumerating validator contents entirely, so it holds whether the validator keeps `eu1` or not, and there is nothing here for the follow-up to delete. The `eu1` removal itself is filed as #165.

docs/api-reference.md
jisung-02: Filed, all six, before merge:

- #165 — reject `eu1` at the validator and drop it from the 141 tool docstrings
- #166 — install `ruff` and `pre-commit` as dev dependencies, drop `black`, `isort`, `flake8`
- #167 — type the shared nonce cookie attributes as a `TypedDict`
- #168 — add `ruff format --check` so Python blocks in markdown stay formatted
- #169 — land the doc-versus-signature comparison as a test or a CI step
- #170 — reconcile the two dev dependency lists in `pyproject.toml`

The Follow-ups section of the PR body now references them by number, so the reasoning and the issue point at each other rather than the reasoning living only on a page that stops being findable.

#165 carries your point verbatim: for an MCP client the tool description is the documentation, so until it lands the guides and the surface the model actually reads disagree. It also records what evidence it lacks — whether an `eu1` region exists on the platform at all is unconfirmed, and that has to be answered before the removal, not after.

junho226 (approved):
## Re-review

The only commit since my last review is `2e0f1df`, which addresses the comment I left on the region note. The fix is the right one: dropping the enumeration and stating what `dev` is leaves a sentence that stays true whether or not #165 lands, instead of one that would have to be edited again.

## Independent verification, re-run

I re-derived the central claim from the source rather than reading the diff, AST-parsing every registration and comparing it against every callable and URI the markdown names.

| Check | Result |
|---|---|
| Registered tools (`@mcp.tool` / `@mcp_tool_handler`) | 166 |
| Resource URIs in `tools/resources.py` | 77 |
| Callables named in docs with no registration | **0** |
| Resource URIs in docs with no code counterpart | **0** |

The seven callables my first pass flagged all resolve:

- `work_session_workflow`, `guarded_execution`, `incident_response`, `security_audit` are **prompts**, not tools, and `docs/api-reference.md:846` classifies them as exactly that — "Prompts (workflow guides, defined in `tools/prompts.py`)". All four signatures match `tools/prompts.py`.
- `validate_file_path` lives in `utils/error_handler.py` and is called inline nine times from `tools/webftp_tools.py`, which is what `CLAUDE.md:58` says about it.
- `error_response` / `success_response` are from `CHANGELOG.md`, outside this branch.

The twenty URI "mismatches" are all the deliberate shorthand — brace alternation (`{cpu|memory|disk-io|...}`) and elision (`alpacon://tokens/...`). My run reproduced precisely the candidate set the PR body says it raised and rejected, which is a good sign for the method as much as for the result.

### Individual claims from the table

- `get_top_servers(workspace, metric_types: str = '', region='')` — comma-separated string, endpoint keys `disk_io` and `traffic`, docs say top five ✓
- `work_session_create` required = `workspace, scopes, servers, expires_at, description` ✓
- `health_check` takes no parameters, so the old "only tool without `workspace`" line was false and is correctly gone ✓
- `execute_command` required = `server_id, command, workspace` ✓
- `main.py:96-111` — a first run with no tokens runs the wizard and returns without starting the server ✓
- `alpacon://server-notes/{region}/{workspace}/{note_id}` is registered at `tools/resources.py:75` and documented at `docs/api-reference.md:834` ✓

### Scope

`git diff --name-only origin/main...HEAD` yields three non-markdown paths: `.dockerignore`, `.gitignore`, `.rules`. No `.py`, no manifest, no lockfile. The no-behavior-change claim holds.

Follow-ups #165 through #170 are all open, so the findings left the PR body and became trackable.

## One thing to watch (not blocking)

`eu1` now disagrees across surfaces. `CLAUDE.md` says regions are `ap1` and `us1` only, while `eu1` still appears **168 times** in `.py` — and for an MCP client the tool docstring *is* the documentation, so the model keeps reading `Region (ap1, us1, eu1)` no matter what the guides say.

The PR body meets this head on and the call is right: the guides should state the truth and #165 should fix the code. I only want to flag that the longer the gap, the longer the drift this PR set out to remove survives in the surface that reaches the model most directly. Worth landing #165 soon.

---

**Verdict**: APPROVED

Every point from my earlier reviews is addressed, and the central claim comes back at zero mismatches on an independent run. What remains all needs a code or manifest change and is filed.

PR #171, opened 2026-08-07
feat(oauth): require S256 PKCE at /oauth/authorize

jisung-02:
## Background

`/oauth/authorize` forwarded an authorization request to Auth0 whether or not it carried a `code_challenge`, while the discovery document already advertised `code_challenge_methods_supported: ['S256']`. A client reading that document got a promise the endpoint did not keep. The proxy now requires S256 PKCE itself instead of leaving the choice to the client.

## What changed

- A request with no `code_challenge` is rejected with `invalid_request` and 400.
- A request that sends a `code_challenge` under any `code_challenge_method` other than `S256` is rejected the same way. `plain` is not accepted.
- A `code_challenge` that is not shaped the way RFC 7636 §4.1 fixes—43 to 128 characters from `[A-Za-z0-9-._~]`—is rejected too.
- The check sits after the `redirect_uri` gate, so a request failing both is told about the `redirect_uri` first.
- `code_challenge_methods_supported` in the discovery document and the gate read the same `_PKCE_CHALLENGE_METHOD` constant, so the enforced and the advertised method cannot drift apart.
- The `/authorize` fallback route delegates to `/oauth/authorize` and inherits the rule with no second copy of it.

## The challenge has to be a challenge

The first version of the gate took `code_challenge` on truthiness alone. A single space, a one-character string, and standard base64 with `+` `/` `=` all cleared it and were forwarded upstream. Auth0 refuses them, so no flow ever completed that way, but a gate whose rejection reads "this server accepts only S256 PKCE" and then accepts one character is not checking what its error message describes. `_PKCE_CHALLENGE_PATTERN` now holds it to the shape, and the failure is local, specific, and testable instead of arriving as a vendor error several redirects later.

The pattern is matched, never normalized. `dict(request.query_params)` is what the gate reads and `urlencode` re-encodes that same dict upstream, so the value checked is the value forwarded. A `.strip()` would break that agreement to accept input the RFC already excludes, so the gate rejects rather than repairs.

Two behaviors that were correct only by accident are now pinned. The method comparison rejects `s256` and a trailing space because it is `!=` against the raw string—right, since RFC 7636 §4.3 makes the method case-sensitive, but a later `.strip().upper()` cleanup would have quietly undone it. And the duplicate-parameter case, which is the one that matters for a proxy: all four orderings agree between the value the gate checks and the value it forwards.

One smaller thing: on the exempt callback, a request with no challenge and `code_challenge_method=garbage` used to reach Auth0 with that value intact. It is inert—Auth0 has nothing to verify without a challenge—but the server declined to inspect it, so it no longer vouches for it upstream either.

## The MFA replay is now load-bearing

The gate sits above the `mfa_requested` branch, so a two-stage flow is rejected without a challenge like any other. That makes the stage-2 replay carry weight it did not carry before: stage 1 authorizes against the MFA audience without the challenge, so the signed state holds the only copy of it. Losing that copy strands the client—it still holds a verifier, and the code it finally exchanges was issued without one.

Nothing pinned that path, so `TestMfaPkceReplay` now does: stage 1 stashes the challenge in the signed state and omits it from the upstream request, stage 2 replays it, and a key outside `_ALLOWED_REPLAY_KEYS` cannot ride the state into stage 2.

## The one exemption

`https://global.consent.azure-apim.net/redirect` — the Power Platform gateway callback a Copilot Studio connector uses. That connector cannot send a challenge, so this destination may start a flow without one. It is narrow:

- The exemption is the exact URI. Another path on the same host is not exempt.
- It goes through `_is_exact_allowed_redirect_uri`, not `_check_redirect_uri`, so report-only mode cannot widen it.
- An exempt destination still may not downgrade to `plain`.

`_log_authorize_client_profile` records what each client actually sends, to settle who still reaches that callback. Remove it once nobody does.

## Verification

- `1127 passed` — 39 new tests: 5 on the exemption predicate, 13 on the gate, 3 on the MFA replay, 10 on the challenge format and the exact method comparison, 4 on duplicate parameters, 4 on the length bounds and rejection ordering.
- `ruff check .` and `ruff format --check .` pass.
- The 20 pre-existing authorize tests now pass `PKCE_PARAMS`, since they have to clear the gate to reach what they assert. `PKCE_PARAMS` carries the challenge from RFC 7636 appendix B, because the shape check refuses the 14-character placeholder it used before.
- `mypy --ignore-missing-imports --no-strict-optional .` reports the same 8 `set_cookie` / `delete_cookie` errors as `main`, shifted by the lines this PR adds. None are new; CI marks that step `continue-on-error: true`.

## Manual check: done, Auth0 enforces the verifier

`/oauth/token` is a pure proxy—it pins `grant_type` and `client_id`, injects `client_secret`, and never inspects `code_verifier`—so the property this PR buys is enforced by Auth0 and no test here can defend it. Run by hand against `mcp.dev.alpacon.io`, two rounds with a fresh authorization code each:

| Round | Exchange | Result |
|---|---|---|
| negative | `code_verifier` omitted | `400` `invalid_request`, "Parameter 'code_verifier' is required" |
| control | `code_verifier` sent | `200`, access, refresh and id tokens issued |

The rejection came from Auth0 and was relayed, and it holds for the confidential-client exchange—`client_secret` injected—which is the case where authorization servers have historically differed. The control rules out the code simply being unredeemable. Full output in the PR thread.

## Review notes

- Whether any other callback belongs in the exemption set. Copilot Studio is the only one today.


tests/test_oauth.py
```diff
@@ -75,6 +79,9 @@
 
 REPORT_ONLY = {'ALPACON_MCP_REDIRECT_URI_REPORT_ONLY': 'true'}
 
+# Every non-exempt authorize request has to carry these
+PKCE_PARAMS = {'code_challenge': 'test-challenge', 'code_challenge_method': 'S256'}
```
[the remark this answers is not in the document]
jisung-02: Fixed in 7cbd3d8. `PKCE_PARAMS` now builds `code_challenge_method` from `_PKCE_CHALLENGE_METHOD`, so the tests and the gate read one value.

utils/oauth.py
```diff
@@ -72,6 +72,11 @@
 # registration from driving a check — and in report-only mode a warning — per entry.
 _MAX_REGISTERED_REDIRECT_URIS = 20
 
+# Both the authorize gate and the discovery document use this, so the enforced
+# and the advertised method cannot drift apart. plain binds nothing an
+# interceptor who already has the authorization request cannot replay.
```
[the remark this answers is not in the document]
jisung-02: Reworded in 7cbd3d8. The original sentence meant "plain binds nothing beyond what an interceptor could already replay", which is the same conclusion, but the double negative made it read as the opposite. It now says it plainly: with `plain` the challenge is the verifier itself, so whoever reads the authorization request can replay it.

utils/oauth.py
```diff
@@ -72,6 +72,12 @@
 # registration from driving a check — and in report-only mode a warning — per entry.
 _MAX_REGISTERED_REDIRECT_URIS = 20
 
+# Both the authorize gate and the discovery document use this, so the enforced
+# and the advertised method cannot drift apart. plain is not accepted: its
+# challenge is the verifier itself, so whoever reads the authorization request
+# can replay it.
+_PKCE_CHALLENGE_METHOD = 'S256'
```
junho226, line 79: Small correction for the PR description rather than for this line: the Background says the discovery document "advertised both `S256` and `plain`", but on `main` `utils/oauth.py:480` is already `'code_challenge_methods_supported': ['S256']`, and `plain` appears nowhere near PKCE in the tree.

The constant is still worth extracting — it is what keeps the advertised and enforced values from drifting later. It just is not removing `plain` from anything, because the real prior gap was the opposite one: S256 advertised, nothing enforced.

utils/oauth.py
```diff
@@ -550,6 +574,30 @@ async def oauth_authorize(request):
...
+        if not code_challenge and not _is_pkce_exempt_redirect_uri(client_redirect_uri):
+            return JSONResponse(
+                {
+                    'error': 'invalid_request',
+                    'error_description': (
+                        'code_challenge is required; this server accepts only '
+                        f'{_PKCE_CHALLENGE_METHOD} PKCE'
+                    ),
+                },
+                status_code=HTTPStatus.BAD_REQUEST,
```
junho226, line 593: (nit, non-blocking) `HTTPStatus.BAD_REQUEST` here and at `:598`, but the `redirect_uri` rejection ten lines up uses `status_code=400` — and the file runs 16 literals to 2 constants. The three gates read as a sequence, so one spelling across them would be easier to scan. Fine to leave until you next touch the file.

junho226 (dismissed):
## What this does

`/oauth/authorize` now requires an S256 `code_challenge` before it forwards anything to Auth0, with one exact-URI exemption for the Power Platform gateway callback. The enforced method and the advertised one are pulled from a single `_PKCE_CHALLENGE_METHOD` constant.

**Approving.** I traced the security property end to end rather than taking the gate at face value, and it holds. Two small things below, neither blocking: one is a factual correction to the PR description, the other is a style inconsistency.

## What I traced, and what held

The gate on its own would be theatre if the challenge did not reach Auth0 and the verifier did not reach it back, so I followed both halves.

- **The challenge reaches Auth0.** `params` is the client's query dict, `code_challenge` / `code_challenge_method` are never stripped, and the standard branch forwards `params` wholesale to `{auth0_base_url}/authorize`. Only `client_id`, `audience`, `scope`, `redirect_uri`, and `state` are rewritten. `test_pkce_challenge_reaches_auth0_unchanged` pins this.
- **The verifier reaches Auth0.** `/oauth/token` parses the client body into `params` and posts it to Auth0 with only `client_id`, `client_secret`, and `redirect_uri` overridden, so a client-sent `code_verifier` passes through untouched. The binding is Auth0's to enforce and nothing here breaks it.
- **The gate covers every entry point.** `/oauth/authorize` and the `/authorize` fallback are the only two routes, and the fallback is a direct `return await oauth_authorize(request)`, so there is no second copy to drift. Confirmed by `test_pkce_fallback_route_enforces_the_same_rule`.
- **The MFA path does not slip past it.** The gate sits above the `mfa_requested` branch, so a two-stage flow is rejected without a challenge like any other. And because the gate now guarantees the challenge exists, the stage-2 replay became load-bearing — that path is intact: `stage2_authorize_params` carries `code_challenge` / `code_challenge_method` into the signed state, and the callback replays them through an explicit `_ALLOWED_REPLAY_KEYS` whitelist with an `isinstance(value, str)` check, so a forged state cannot inject arbitrary authorize params.
- **The exemption is not a bypass anyone can claim usefully.** Any client may send `redirect_uri=https://global.consent.azure-apim.net/redirect` and skip the challenge, but the resulting code is relayed to that URI, not to the caller, so claiming the exemption buys an attacker nothing. `_is_pkce_exempt_redirect_uri` requiring `_is_exact_allowed_redirect_uri` rather than `_check_redirect_uri` is the right call: it keeps `ALPACON_MCP_REDIRECT_URI_REPORT_ONLY` from widening the exemption to any path on that host, and `test_pkce_exemption_is_not_widened_by_report_only_mode` pins it.
- **The empty-string cases behave.** `code_challenge=` is falsy and rejected; a challenge with no method is rejected, which is stricter than RFC 7636's `plain` default and correct here; a missing `redirect_uri` skips the redirect gate but still requires PKCE, since `''` is not in the exempt set.

The test set matches the surface. 18 new tests: 5 unit tests on the exemption predicate and 13 on the gate, covering missing challenge, empty challenge, missing method, `plain`, S256 accepted, exempt-may-omit, exempt-may-not-downgrade, deeper path on the exempt host, no redirect_uri, gate ordering, forwarding, the fallback route, and advertised-equals-enforced. Every case I probed by hand was already there.

## 1. (low) The Background misstates the prior state

> the discovery document advertised both `S256` and `plain`

It did not. On `main`, `utils/oauth.py:480` reads `'code_challenge_methods_supported': ['S256']`, and `plain` appears nowhere near PKCE anywhere in the tree. The diff bears this out — the line changes from `['S256']` to `[_PKCE_CHALLENGE_METHOD]`, which is the same value behind a constant.

Worth correcting, because the accurate version is the better argument. The gap was not that the server advertised a weak method; it was that it advertised S256 and then enforced nothing, so a client reading the discovery document got a promise the endpoint did not keep. Extracting the constant does not remove `plain` from anything — it stops the advertised value and the enforced value from drifting apart later, which is what the PR says elsewhere and is worth saying only once.

## 2. (nit) `HTTPStatus.BAD_REQUEST` next to a literal `400`

The two new rejections use `status_code=HTTPStatus.BAD_REQUEST`; the `redirect_uri` rejection ten lines above uses `status_code=400`. The file is 16 literals to 2 constants, so the new code is the minority spelling in its own function. Either is fine, but the two gates read as a pair and should look like one. Not worth a round trip on its own — fold it in if you touch the file again.

## Checks run

Worktree at `3dd3fdb`, the PR head. Commands are the ones `.github/workflows/test.yml` runs.

| Command | Result |
|---|---|
| `PYTHONPATH=. uv run --no-sync pytest tests/ -q --tb=short` | **`1106 passed, 8 warnings in 7.98s`** — matches the number in the description exactly |
| `uv run --no-sync ruff check .` | `All checks passed!` |
| `uv run --no-sync ruff format --check .` | `82 files already formatted` |
| `uv run --no-sync mypy --ignore-missing-imports --no-strict-optional .` | `Found 8 errors in 1 file (checked 82 source files)` |

The 8 mypy errors are the pre-existing `set_cookie` / `delete_cookie` `**dict[str, object]` complaints. I checked them out on `main` to be sure this PR did not add any: identical eight, at `utils/oauth.py:422` / `:437` there and `:446` / `:461` here — the same errors moved down by the lines this PR adds. CI marks that step `continue-on-error: true`, so it does not gate either way.

I also re-derived the two count claims in the description rather than trusting them: 18 new test functions in `tests/test_oauth.py`, splitting 5 predicate / 13 gate as stated.

**Not run:** CI's Python 3.13 and 3.14 matrix legs — the local venv is 3.12.9, so only the 3.12 leg was exercised. Also the coverage gate (`--cov-fail-under=60`); I ran pytest without `--cov`, so I cannot report the number CI computes.

**Not verifiable here, and it is the one that matters most:** the manual check in the description — completing a real authorization against `mcp.dev.alpacon.io` and replaying the token exchange with `code_verifier` omitted to confirm Auth0 rejects it. That is Auth0's half of the contract and needs a live tenant. Everything above establishes that this code puts the challenge in front of Auth0 and the verifier back to it; it does not establish what Auth0 then does with them. Worth doing before merge as you planned.

## One thought on the review note you left open

On whether other callbacks belong in the exemption set: `_log_authorize_client_profile` logs `redirect_uri` alongside the method on every authorize request, so the data to answer that is already being collected and the empty-method case reads as the no-challenge one. Leaving the set at one entry until the logs say otherwise is the right default — every addition is a destination that can start a flow with no replay protection.

utils/oauth.py
jisung-02: Corrected in the PR description. The Background now says what the diff actually shows: `main` already advertised `code_challenge_methods_supported: ['S256']` and enforced nothing, so a client reading the discovery document got a promise the endpoint did not keep. Dropped the claim that `plain` was ever advertised, and left the drift argument where it already was rather than repeating it.

utils/oauth.py
jisung-02: Agreed on the direction — the new lines are the minority spelling, not the literal ten lines up. Leaving it for now and folding it in the next time this file is touched, so this PR stays PKCE-only.

jisung-02:
Pushed `ca52995`, one test-only commit, and moved the unrelated doc corrections out to `docs/fix-stale-tool-references` so this PR stays PKCE-only.

**What the commit adds.** `TestMfaPkceReplay` in `tests/test_oauth.py`, three tests, no production code. It covers the observation in your review that the stage-2 replay became load-bearing once the gate started guaranteeing the challenge exists. `authorize_params` appeared zero times in `tests/test_oauth.py` before this, so that path was traced by hand in review and by nothing else:

- `test_stage1_stashes_the_challenge_in_the_state` — stage 1 omits `code_challenge` from the upstream request and the signed state carries both the challenge and the method.
- `test_stage2_replays_the_stashed_challenge` — the callback puts both back on the stage-2 authorize request.
- `test_stage2_replays_only_the_allowlisted_keys` — an `audience` planted in `authorize_params` does not survive `_ALLOWED_REPLAY_KEYS`; the server value wins.

Each was checked against a mutation of `utils/oauth.py` that only it catches.

`1109 passed` locally on 3.12, `ruff check .` and `ruff format --check .` clean. The PR description is updated to match: the count, and a section on why the replay carries weight now.

The two items from your review: the Background correction is in, and the `HTTPStatus.BAD_REQUEST` / `400` split is deferred to the next change that touches this file.

junho226 (commented):
## Summary

This closes a real gap: the discovery document advertised `code_challenge_methods_supported: ['S256']` while `/oauth/authorize` forwarded anything. The gate is sound on every downgrade path I probed, and the challenge reaches Auth0 unchanged, so the binding to the authorization code happens where it should—at the authorization server that issues the code.

I drove the added gate directly against the real route rather than reading it, covering the RFC 7636 downgrade paths plus query-string parameter pollution. Results are in **Verification** below. Two findings stay open, neither of them a bypass.

## Critical issues

None. I could not construct a request that reaches Auth0 with a weaker challenge than the gate claims to allow.

## Warnings

**1. The verifier is never checked in this repo—the gate's value rests entirely on Auth0.**

`/oauth/token` (`utils/oauth.py:670`) is a pure proxy: it pins `grant_type` and `client_id`, injects `client_secret`, and forwards the body to Auth0. It never inspects `code_verifier`. That is the right architecture—this server does not issue codes, so it cannot hold the challenge—but it means the property this PR is actually buying ("a stolen code cannot be redeemed without the verifier") is enforced by a system outside this repository, and nothing here proves it.

The sharp edge is the `client_secret` injection. The exchange Auth0 sees is a *confidential-client* exchange, and PKCE verification for a confidential client that authenticates with a secret is exactly the case where authorization servers have historically differed. Auth0 does enforce the verifier when the authorization request carried a challenge, so I expect this to hold—but "I expect" is not verification, and I have no Auth0 access from here to settle it.

The PR already schedules this as a manual check against `mcp.dev.alpacon.io`. That is the right call. Please treat it as blocking-before-merge rather than nice-to-have, and record the result in the PR thread—it is the one assertion in this change that no test in the repo can defend.

**2. The gate checks that a challenge is present, not that it is a challenge.**

`code_challenge` is accepted on truthiness alone. RFC 7636 §4.1 fixes the shape: 43–128 characters from the unreserved set `[A-Za-z0-9-._~]`. Measured against the live route:

| Sent | Result |
|---|---|
| `code_challenge=%20` (a single space) | **302**, forwarded upstream as `' '` |
| `code_challenge=x` (1 char, RFC requires 43–128) | **302**, forwarded as `x` |
| `code_challenge=a%2Bb%2Fc%3D` (`+`, `/`, `=`—standard base64, not base64url) | **302**, forwarded as `a+b/c=` |

Auth0 rejects a malformed challenge downstream, so no request completes an authorization this way. But a gate whose stated job is "this server accepts only S256 PKCE" and which accepts a one-character challenge is not checking the thing its error message describes. `.strip()` plus a length-and-charset check is a few lines and makes the rejection local, specific, and testable instead of deferred to a vendor error the client sees later.

## Suggestions

**3. Note in the docs that the MFA first leg carries no challenge.** Confirmed against the route: with `scope=openid mfa`, stage 1 redirects to the MFA audience with `code_challenge` **absent** upstream—the challenge is stashed in the signed state and replayed at stage 2 (`utils/oauth.py:625-642`, `1110-1123`). That is fine: the stage-1 code is exchanged server-to-server with `client_secret` and discarded (`utils/oauth.py:1065-1090`), so client authentication is what protects it. But the new line in `docs/configuration.md` reads as an unqualified "`/oauth/authorize` requires S256 PKCE", and a future reader tracing the MFA path will find a leg that does not carry one. Half a sentence prevents that.

**4. Pin the cases that currently pass by accident of framework semantics.** All three below behave correctly today—I checked—but nothing in the suite holds them there:

- `code_challenge_method=s256` and `S256 ` (trailing space) → correctly 400, but only because the comparison is `!=` on an untrimmed string. A later `.strip().upper()` "cleanup" would silently accept both.
- Duplicate query parameters. This is the one that matters for a proxy: the value the gate checks must be the value forwarded upstream. `dict(request.query_params)` is last-wins and `urlencode(params)` re-encodes that same dict, so they agree—`?code_challenge=&code_challenge=abc` is checked as `abc` and forwarded as `abc`, and `?code_challenge_method=plain&code_challenge_method=S256` likewise. That agreement is load-bearing and incidental. A test costs three lines and catches any future move to `getlist` or a framework with first-wins semantics.

**5. Drop `code_challenge_method` when there is no challenge.** On the exempt URI, `?code_challenge_method=garbage` with no challenge is forwarded upstream verbatim. The PR body calls this the intended width and it is inert—Auth0 has nothing to verify—but forwarding a value the server declined to inspect is avoidable.

The `HTTPStatus.BAD_REQUEST` / bare `400` split is already settled in the thread as deferred; not re-raising it.

## Highlights

- **The exemption is genuinely narrow, and I verified each claim rather than taking the docstring's word.** Routing `_is_pkce_exempt_redirect_uri` through `_is_exact_allowed_redirect_uri` instead of `_check_redirect_uri` holds up: with `ALPACON_MCP_REDIRECT_URI_REPORT_ONLY=true` the exemption does not widen, a deeper path on the same host is rejected by the redirect gate first (400), and dropping the URI from `ALLOWED_REDIRECT_URIS` kills the exemption with it (predicate `False`, authorize 400). Tying an exemption's lifetime to the allowlist entry it depends on is the detail that keeps this from rotting into a permanent hole.
- **A missing `code_challenge_method` is rejected, not defaulted to `plain`.** RFC 7636 §4.3 says the default is `plain`; following that default here would have silently reintroduced the downgrade. Choosing the safe reading over the literal one is the right call.
- **`_PKCE_CHALLENGE_METHOD` is shared between the gate and the discovery document, with a test pinning them together.** The bug this PR fixes was advertised-versus-enforced drift; the fix makes that class of bug unrepresentable rather than just correcting today's instance.
- **Rejections are a JSON 400, never an error redirect.** An unvalidated `redirect_uri` receives nothing, and the redirect gate runs first regardless.
- **`TestMfaPkceReplay` pins the replay-key allowlist**, so a forged state cannot ride arbitrary parameters into the stage-2 upstream request. Adding it in response to review, on a path that had zero coverage, is the right instinct.

## Verification

Run in a worktree at `ca52995`, the PR head.

| Command | Result |
|---|---|
| `uv run ruff check .` | pass—`All checks passed!` |
| `uv run ruff format --check .` | pass—`82 files already formatted` |
| `uv run pytest tests/test_oauth.py -q` | pass—`173 passed` |
| `uv run pytest -q` (full suite) | pass—`1109 passed, 7 warnings in 7.04s`, matching the PR's stated count |
| Ad-hoc probe of the authorize route (not committed) | see below |

The probe drove `/oauth/authorize` through Starlette's `TestClient` against the real registered routes, asserting on both the status code and the parameters forwarded upstream:

- **Rejected (400 `invalid_request`), as intended:** no challenge; empty challenge; challenge with no method; `plain`; `s256`; `S256 ` with a trailing space; no `redirect_uri` and no challenge; the `/authorize` fallback route with no challenge; a deeper path on the exempt host.
- **Accepted, correctly:** `S256` with a challenge, forwarded unchanged; the exempt URI omitting the challenge, with no `code_challenge` sent upstream.
- **Parameter pollution:** all four duplicate-parameter orderings agree between the checked value and the forwarded value.
- **Accepted, and the basis of warning 2:** whitespace-only, 1-character, and non-base64url challenges.
- **MFA stage 1:** redirects to the MFA audience with `code_challenge` absent upstream (basis of suggestion 3).

**Could not run:** the end-to-end check against Auth0—that a code issued with a challenge is refused at `/oauth/token` when `code_verifier` is omitted. It needs a live Auth0 tenant and credentials I do not have from here. This is warning 1, and it is the PR's own pre-merge manual check.

## Verdict

**Comment.** The gate holds, the exemption is narrow and self-limiting, and the test coverage of the new behavior is genuinely good—21 tests that are mostly negative cases, which is the right shape for a security gate. Nothing here blocks merge on correctness.

Two things stay open: the challenge-format validation in warning 2, and confirming the Auth0-side verifier check in warning 1 before this ships. The second is the one that matters—it is the assumption the whole change rests on, and it is the only claim no test in this repository can defend.

## Checklist

- [x] `code_challenge` required at `/oauth/authorize`, `plain` rejected
- [x] `code_challenge_method` restricted to exactly `S256`; case-sensitive, whitespace-sensitive
- [x] Missing method rejected rather than defaulted to `plain` per RFC 7636 §4.3
- [x] Challenge forwarded to the authorization server unchanged, so it binds to the code
- [ ] Challenge format validated per RFC 7636 §4.1 (43–128 chars, unreserved charset)—**not validated**
- [x] Rejection precedes any redirect; unvalidated `redirect_uri` never receives an error redirect
- [x] Fallback `/authorize` route inherits the gate
- [x] Exemption is exact-match, not widened by report-only mode, and dies with its allowlist entry
- [ ] Verifier checked at the token endpoint—**delegated to Auth0, unverified**
- [x] Negative test coverage: missing, empty, `plain`, missing method, exemption scope, MFA replay keys
- [ ] Negative test coverage: case/whitespace variants, duplicate parameters—**missing, behavior currently correct**
- [x] Backward compatibility understood and documented: existing clients that send no challenge break by design, with one exempt callback

utils/oauth.py
```diff
@@ -550,6 +574,30 @@ async def oauth_authorize(request):
                 status_code=400,
             )
 
+        code_challenge = params.get('code_challenge', '')
```
junho226, line 582: Warning: the gate checks that a challenge is present, not that it is a challenge.

`code_challenge` is accepted on truthiness alone, so three inputs that are not valid challenges under RFC 7636 §4.1 (43–128 chars from `[A-Za-z0-9-._~]`) get a 302 and are forwarded upstream. Measured against this route at `ca52995`:

- `code_challenge=%20` (a single space) → 302, forwarded as `' '`
- `code_challenge=x` (1 char) → 302, forwarded as `x`
- `code_challenge=a%2Bb%2Fc%3D` (`+`, `/`, `=` — standard base64, not base64url) → 302, forwarded as `a+b/c=`

Auth0 rejects a malformed challenge downstream, so nothing completes an authorization this way. But the error message one line down promises "this server accepts only S256 PKCE", and a one-character challenge is not that. `.strip()` plus a length-and-charset check keeps the rejection local, specific, and testable instead of surfacing later as a vendor error the client has to interpret.

utils/oauth.py
```diff
@@ -550,6 +574,30 @@ async def oauth_authorize(request):
...
+                {
+                    'error': 'invalid_request',
+                    'error_description': (
+                        'code_challenge is required; this server accepts only '
+                        f'{_PKCE_CHALLENGE_METHOD} PKCE'
+                    ),
+                },
+                status_code=HTTPStatus.BAD_REQUEST,
+            )
+        if code_challenge and code_challenge_method != _PKCE_CHALLENGE_METHOD:
```
junho226, line 595: Suggestion: this comparison is correct today for reasons worth pinning with a test.

I verified that `s256` and `S256 ` (trailing space) both get a 400 — but only because the comparison is `!=` against an untrimmed, case-sensitive string. RFC 7636 §4.3 makes the method case-sensitive, so that is right; a later `.strip().upper()` "cleanup" would silently start accepting both. Nothing in the suite holds that line.

The duplicate-parameter case matters more for a proxy: the value this gate checks has to be the value forwarded upstream. `dict(request.query_params)` is last-wins and `urlencode(params)` re-encodes that same dict, so they agree — `?code_challenge_method=plain&code_challenge_method=S256` is checked as `S256` and forwarded as `S256`, and the reverse ordering gets a 400. That agreement is load-bearing and incidental to the framework. A three-line test would catch a future move to `getlist` or a first-wins framework.

utils/oauth.py
jisung-02: Fixed in 804c829. `_PKCE_CHALLENGE_PATTERN` holds the challenge to the shape RFC 7636 §4.1 fixes, and all three inputs you measured now get a 400 with `code_challenge must be 43 to 128 characters from the RFC 7636 unreserved set`.

One deliberate departure from your suggestion: no `.strip()`. Trimming would decouple the value the gate checks from the value it forwards, which is exactly the agreement you called load-bearing in the duplicate-parameter note below. So the pattern rejects rather than repairs — whitespace fails the charset anyway, and a leading or trailing space is not something the RFC permits a client to send. The reason is in the constant's comment so a later reader does not add the `.strip()` back.

`TestAuthorizePkceChallengeFormat` covers your three cases plus `'A' * 42`, `'A' * 129`, and a trailing newline, with `43` and `128` pinned as accepted so the bounds stay inclusive. The 129 and 42 cases are what would catch an off-by-one if someone rewrites the quantifier.

utils/oauth.py
jisung-02: Pinned in 5c25947. `test_pkce_method_comparison_is_exact` parametrizes `s256`, `S256 `, ` S256`, and `SHA256`; its docstring says why they pass, so the next person to reach for `.strip().upper()` sees the test fail and reads the reason rather than deleting it.

The duplicate-parameter case is in `TestAuthorizePkceDuplicateParams`, four tests. For `code_challenge` and `code_challenge_method` alike, one test asserts that the last value is both what the gate accepts and what lands in the forwarded query string, and one asserts that reversing the order makes the same request fail. That is the part your note was really about: it is not enough to know the gate reads last-wins, the forwarded value has to be the same one. The class docstring records that `dict(request.query_params)` and `urlencode` agree incidentally, and names `getlist` and first-wins semantics as the two ways that could break.

Also took the deferred `HTTPStatus` nit in 35e08c0 since the file was open — the `redirect_uri` rejection now spells the status the same way the two gates below it do.

jisung-02:
Five commits on top of `ca52995`, addressing everything from the last review except warning 1, which is not fixable in code.

| Commit | Item |
|---|---|
| `35e08c0` | the deferred `HTTPStatus` nit |
| `804c829` | warning 2 — challenge format |
| `785456c` | suggestion 5 — uninspected method |
| `5c25947` | suggestion 4 — case, whitespace, duplicate parameters |
| `2316228` | suggestion 3 — MFA first leg in the docs |

Warning 2 and suggestion 4 are answered in their own threads. The rest:

**Suggestion 5.** `params.pop('code_challenge_method', None)` when there is no challenge, so the exempt callback no longer carries `code_challenge_method=garbage` upstream. The pop runs after `_log_authorize_client_profile`, which still records what the client actually sent — that log is how we will eventually answer whether anyone besides Copilot Studio reaches the exempt callback, so it has to keep seeing the raw value.

**Suggestion 3.** Half a sentence turned out to be a whole bullet, because the useful part is not "the first leg carries no challenge" but why that is not a hole: the stage-1 code is exchanged server-to-server with `AUTH0_CLIENT_SECRET` and discarded, so client authentication protects it, and the client's own challenge rides the signed state to the second leg — the one whose code the client actually exchanges. I re-read `utils/oauth.py:1084-1113` rather than take the shape of it from your note.

**Warning 1.** Agreed, and it is now in the PR body as blocking rather than as a checklist item, with your framing of why: Auth0 sees a confidential-client exchange because of the `client_secret` injection, and that is the case where authorization servers have historically differed on whether the verifier is still checked. I will run it against `mcp.dev.alpacon.io` and post the result here before merge.

One more thing worth flagging, since it moved: the eight `mypy` errors are not new, but they are not absent from `main` either. I checked out `origin/main` in a scratch worktree and got the identical eight at `utils/oauth.py:422` / `:437`, against the same `starlette 1.3.1` and `mypy 1.18.2` this branch uses. Your read was right; I am recording the check because the older clone I first compared against was several commits behind and came back clean, which was misleading.

`1127 passed`, `ruff check .` and `ruff format --check .` clean.

jisung-02:
## Warning 1 settled: Auth0 does enforce the verifier

Ran the pre-merge check against `mcp.dev.alpacon.io`. Two rounds through the real proxy, one browser login, a fresh authorization code each round.

**Round 1, negative** — authorize with an S256 challenge, then exchange with `code_verifier` omitted:

```
HTTP 400
{
  "error": "invalid_request",
  "error_description": "Parameter 'code_verifier' is required"
}
```

**Round 2, control** — authorize with a fresh challenge, then exchange with the verifier:

```
HTTP 200
{
  "access_token": "<1267 chars>",
  "refresh_token": "<90 chars>",
  "id_token": "<1066 chars>",
  "scope": "openid profile email offline_access",
  "expires_in": 300,
  "token_type": "Bearer"
}
```

The control matters: a rejection on its own would only show the exchange failed, not that it failed for the missing verifier. With the same flow succeeding when the verifier is present, the difference is attributable.

Two details worth recording.

**The error is Auth0's, not ours.** `/oauth/token` never reads `code_verifier`—the string appears once in `utils/oauth.py`, in a comment—so that 400 was produced upstream and relayed. And it is `invalid_request` / "Parameter 'code_verifier' is required", not `invalid_grant`: Auth0 is not failing a comparison, it is refusing an exchange it considers incomplete because the authorization request carried a challenge.

**It holds for the confidential-client case you flagged.** The exchange Auth0 saw had `client_secret` injected, which is exactly where authorization servers have historically differed on whether PKCE is still verified. On this tenant it is. That was the assumption the whole change rests on, and it is no longer an assumption.

Round 1 used a code that Auth0 issued against a challenge and then declined to redeem without the verifier—which is the stolen-code scenario the gate exists for, minus the theft.

Both authorization codes are spent. The check script is not committed; it needs a browser and a live tenant, so it does not belong in the suite—which is the same reason this could not be a test and had to be run by hand.

junho226 (approved):
## Round 2

Five commits since `ca52995`. Every finding from the previous round is closed, and the one that needed a live tenant was closed with evidence rather than an assertion.

| Prior finding | Status |
|---|---|
| W1 — `/oauth/token` never inspects `code_verifier`; the property rests on Auth0 enforcing RFC 7636 §4.6 for a confidential-client exchange | **Settled with evidence.** Not code — the pre-merge check against `mcp.dev.alpacon.io`, negative plus control |
| W2 — the gate checked presence, not shape | **Fixed** (`804c829`) |
| S3 — MFA stage 1 reaches Auth0 with no challenge, undocumented | **Fixed** (`2316228`) |
| S4 — exact method comparison and duplicate-parameter handling unpinned | **Fixed** (`5c25947`) |
| S5 — exempt URI forwarded an uninspected `code_challenge_method` | **Fixed** (`785456c`) |

### W1 — the part I could not run, and why I accept it

This was the one gap I flagged last round and could not close from here. The check you ran does close it, and the methodology is why: the negative round alone would only show the exchange failed, and the control round makes the failure attributable to the missing verifier. Two further details carry the weight — the 400 is `invalid_request` / "Parameter 'code_verifier' is required", which is Auth0 refusing an incomplete exchange rather than failing a comparison, and the exchange carried `client_secret`, which is precisely the confidential-client case where authorization servers have historically differed. That was the assumption the change rested on and it is now measured.

I did not re-run this myself — it needs a live tenant and a browser, neither of which I have here. I am accepting your transcript, and recording that distinction rather than blurring it.

### W2 — fixed, and the anchor choice is the part worth noticing

`_PKCE_CHALLENGE_PATTERN` rejects all three inputs I measured last round, plus every adversarial case I could construct. Driven against the live route at `2316228`:

| Input | Result |
|---|---|
| 43 and 128 chars unreserved | 302, forwarded unchanged |
| 42 chars / 129 chars | 400 |
| single char / whitespace-only / embedded space | 400 |
| `+`, `/`, `=` (standard base64, not base64url) | 400 |
| `-`, `.`, `_`, `~` | 302 — the unreserved set is complete |
| valid challenge + trailing newline | 400 |

That last row is the one I want to name. The pattern ends `\Z`, not `$`. In Python `$` also matches immediately before a trailing newline, so the `$` spelling would have accepted `<43 valid chars>\n` — I confirmed the counter-case directly: the real pattern rejects it, the same pattern with `$` accepts it. `tests/test_oauth.py:1107` pins it with `f'{PKCE_CHALLENGE}\n'`, so the anchor cannot be quietly downgraded. This is the kind of detail that is invisible until it is wrong.

Declining to `.strip()` is also right, and for the stated reason: trimming would decouple the value the gate checks from the value it forwards, which is the property the whole gate depends on.

### S4 — the agreement is now pinned, not incidental

`test_pkce_method_comparison_is_exact` parametrizes `s256`, `S256 `, ` S256`, and `SHA256`, and the docstring explains why they must fail. That is the right shape for this test: someone reaching for `.strip().upper()` gets a failure that explains itself instead of one they delete.

`TestAuthorizePkceDuplicateParams` covers the case I could not otherwise have relied on. I re-measured all four orderings against the live route and the checked value is the forwarded value in each — valid-then-bad gets a 400, bad-then-valid gets a 302 forwarding the valid one. The behavior was correct before, but it was incidental to Starlette's last-wins `dict()` conversion; now it is stated.

### S5 — confirmed absent, not merely unchecked

All three exempt-URI cases now forward with no `code_challenge_method` at all: absent, `garbage`, and `plain` each produce a 302 with the parameter stripped. The comment above the `pop` says the right thing — forwarding it would hand Auth0 a value this server never stood behind.

## New findings

None. I looked for a way past the widened gate and did not find one.

## Highlights

- The rejection order is correct and load-bearing: `redirect_uri` is validated first, so a request with an unvalidated `redirect_uri` never receives an error redirect. The PKCE gates then run presence → method → shape, and `test_pkce_method_is_rejected_before_the_challenge_format` pins that ordering rather than leaving it to reading order.
- `_is_pkce_exempt_redirect_uri` going through `_is_exact_allowed_redirect_uri` instead of `_check_redirect_uri` remains the right call — report-only mode accepts any path on an allowlisted host, and routing the exemption through it would let an environment variable widen the set of destinations allowed to skip PKCE.
- The docs paragraph on the MFA first leg explains why the first leg carries no challenge and what protects it instead (client authentication, server-to-server, code discarded). A reader who finds that request in a log now has the answer without reading the handler.

## Verification

Worktree at `2316228`, the PR head.

| Command | Result |
|---|---|
| `uv run ruff check .` | pass — `All checks passed!` |
| `uv run ruff format . --check` | pass — 82 files already formatted |
| `uv run pytest -q` | pass — **1127 passed**, 7 warnings, 7.40s (up from 1109 last round) |
| Ad-hoc probe against the live registered routes | ran clean — the tables above are its output |

The probe drove `/oauth/authorize` through a Starlette `TestClient` over the real routes from `register_oauth_routes`, asserting on both the status code and the parameters actually forwarded upstream: 14 shape cases against the compiled pattern, 12 cases through the route, 4 duplicate-parameter orderings, and the 3 exempt-URI cases. It lives in a scratch directory, never in the repository.

**Could not run:** the end-to-end Auth0 exchange (W1). It needs a live tenant and an interactive browser login. You ran it and posted both rounds; I am relying on that, as noted above.

## Verdict

**APPROVE.** Every prior finding is closed, four of them in code with tests that pin the reason rather than just the result, and the fifth by a measurement with a control. The gate holds on every downgrade path I could drive at it.

## Checklist

- [x] Tests included and passing
- [x] No security vulnerabilities found
- [x] Documentation updated

PR #173, opened 2026-08-11
fix(tools): steer agents away from sudo prefixes and command-list descriptions

### LINKED ISSUE #172: fix(tools): steer agents away from sudo-prefixed commands and command-list descriptions
## Problem

Reported from Codex sessions operating Alpacon via MCP (internal feedback, 2026-08-10). Two recurring misuse patterns:

1. Agents paste the sudo commands they intend to run into the Work Session `description` instead of executing them as commands via `execute_command`.
2. Agents habitually prefix commands with `sudo` even when root is not required, sending every such call into HITL approval and blocking the session on a human for no reason.

## Root cause

Both patterns are induced by the current tool schemas:

- The `work_session_create` description says only "description: what you are doing and why. Be specific" (`tools/work_session_tools.py:24`). Nothing states that the description is not an execution channel. The `SUDO_INTENT_DEVIATION` hint ("update `work_session_update(description=...)` and re-run", `tools/command_tools.py:53`) reinforces the wrong lesson: putting the command text into the description looks like the way to get sudo approved.
- The `execute_command` tool description (`tools/command_tools.py:277`) never mentions the cost of sudo. The "every sudo invocation routes to HITL" warning lives only in `tools/prompts.py`, but MCP prompts are opt-in—clients like Codex only see tool schemas, so the guidance never reaches them.

## Proposed changes

Schema/description text only; no server-side changes.

1. **`work_session_create` / `work_session_update`**: add to the tool description that `description` is a human-readable purpose statement for the approver, NOT a command list—nothing in it is executed; commands run via `execute_command` once the session is active.
2. **`execute_command` / `execute_command_multi_server`**: add a sudo rule—do not prefix with `sudo` by default; every sudo invocation routes to human-in-the-loop approval and blocks; use sudo only when the command genuinely requires root and the session carries the `sudo` scope. `execute_command` is `anthropic/alwaysLoad: True`, so this is the highest-leverage spot.
3. **`SUDO_INTENT_DEVIATION` hint** (`tools/command_tools.py:50-54`): reword so "re-declare the session's purpose" cannot be read as "embed the command text in the description".
4. **`tools/prompts.py`**: mirror both rules in `work_session_workflow` and `guarded_execution` for clients that do load prompts.

Deliberately out of scope for now: a soft guard that warns when a description looks like a command block. Add only if the schema guidance alone doesn't fix the behavior.

## Verification

Re-run the same scenario in a Codex session after the change and confirm (a) descriptions stay prose-only, (b) unprompted `sudo` prefixes disappear.

jisung-02:
## Background

Codex sessions operating Alpacon over MCP produced two recurring misuses, both induced by the tool surface rather than by the server.

Agents pasted the sudo commands they intended to run into the Work Session `description`, instead of executing them through `execute_command`. The schema invited it: `work_session_create` described the field as "what you are doing and why. Be specific" and never said it is inert. The `SUDO_INTENT_DEVIATION` denial guidance made it worse by telling the agent to "re-declare what the session is for with `work_session_update(description=...)`", which reads as an instruction to embed the command text in the field the approver reads.

Agents also prefixed commands with `sudo` by habit, on commands that do not need root. Unless a Work Session sudo policy already covers the command, each of those either stalls the session on a human approver or is denied outright. The warning existed, but only in `tools/prompts.py`, and MCP prompts are opt-in—a client that reads tool schemas alone never saw it.

## Changes

Description text only. No `.py` control flow, no request shape, and no server-side behavior is touched.

| Surface | Change |
|---|---|
| `work_session_create`, `work_session_update` (`tools/work_session_tools.py`) | `description` is stated as prose for the human approver, not a command list; nothing in it is executed |
| `execute_command`, `execute_command_multi_server` (`tools/command_tools.py`) | the sudo rule enters the tool description, where every client sees it |
| `SUDO_INTENT_DEVIATION` hint (`tools/command_tools.py:49`) | "re-declare" becomes "restate in prose", with the prohibition spelled out |
| `_NEXT_ACTION_BY_CATEGORY['SUDO_INTENT_DEVIATION']` (`utils/common.py:79`) | same correction on the structured `next_action` channel, which the issue did not name but which carried the identical wording |
| `work_session_workflow`, `guarded_execution` (`tools/prompts.py`) | both rules mirrored for clients that do load prompts; em dashes normalized to the unspaced form CLAUDE.md specifies |
| `execute_command`, `execute_command_multi_server`, `work_session_create`, `work_session_update` (`docs/api-reference.md`) | the same two rules in the catalog CLAUDE.md sends a reader to before calling an unfamiliar tool; `sudo` added to the agent-requestable scopes, which that file listed as `command`/`webftp`/`tunnel` while `work_session_workflow` said otherwise |

`execute_command` carries `anthropic/alwaysLoad: True`, so its description is the one place the guidance is guaranteed to be in context.

The sudo claim is qualified rather than absolute. "Every sudo invocation blocks until a human acts" is not true in either direction: a Work Session sudo policy with `allow_bypass_mfa` runs the command with no approval wait (`_SUDO_DENIAL_HINTS['SUDO_POLICY_MFA_REQUIRED']`), and a hard denial such as `SUDO_RISK_DENIED` never becomes something a human can unblock. `execute_command` therefore reads "unless a Work Session sudo policy already covers the command, a sudo invocation either routes to human-in-the-loop approval and blocks until a human acts, or is denied outright with no request anyone can approve." `execute_command_multi_server` says "stalls that server's command until a human acts—once per server" instead: it submits without waiting for results, so "blocks" would read as the call itself hanging. The prompts carry the same qualifier, and `guarded_execution` names `SUDO_RISK_DENIED` as the case where waiting never ends. That keeps the deterrent without sending an agent to wait on a request that was permanently denied.

Both descriptions then name the shape the response actually carries. `execute_command` reads "Check sudo_denial.category before waiting; a sudo_hint with no sudo_denial is a hard denial that creates no request, so do not wait on it." An earlier draft said "Read the returned category before waiting", which pointed at a key no response has. `_attach_sudo_denial` (`tools/command_tools.py:147`) always sets `sudo_hint` and adds the structured object only for codes in `_SUDO_HUMAN_APPROVAL_CODES`, nested as `sudo_denial.category`; `SUDO_RISK_DENIED` is deliberately outside that set, so the hard denial the sentence exists for arrives as a bare `sudo_hint`. `tools/prompts.py` step 4 named the fields correctly, but prompts being opt-in is this PR's own reason for moving the rule into the description, so the description cannot lean on the prompt.

`execute_command_multi_server` sends the reader one hop further, to "that server's list_commands entry", because those two fields never appear on its own response. `_attach_sudo_denial` runs on `list_commands` entries and on `execute_command`, and this tool fills `deploy_results[sid]` straight from the submit call, whose result carries no output yet. The description already ended with "use list_commands to check status", but that is ~130 characters after the sentence naming the field, and a client reading top to bottom looks for it on the value in hand.

Deliberately not done, per the issue: a soft guard that rejects a description looking like a command block. The schema wording is the cheaper first attempt, and a heuristic that misfires on a legitimate description costs more than it saves.

Also out of scope, raised in review and filed as #175: no tool in this server carries parameter-level descriptions, so every per-argument rule here sits in the tool description rather than on the argument a client renders at fill time.

## Testing

- `uv run pytest -q`: `1092 passed, 7 warnings in 7.32s`
- `uvx ruff format --check tools/ utils/ tests/`: `77 files already formatted`; `uvx ruff check tools/ utils/ tests/`: `All checks passed!`
- Review of `af51ead` ran the coverage gate this branch's own runs skipped: `pytest tests/ --cov=. --cov-fail-under=60` gives `Required test coverage of 60% reached. Total coverage: 94.85%`.
- Two regression classes were added, both asserting against the descriptions `mcp.list_tools()` actually returns rather than the source strings: `TestSudoGuidanceInToolDescriptions` (`tests/test_command_tools.py`) and `TestDescriptionIsNotAnExecutionChannel` (`tests/test_work_session_tools.py`). The first pins the policy carve-out, the hard-denial clause, and `sudo_denial.category` by substring, so the wording those three review rounds produced cannot regress silently.

The behavioral verification the issue asks for—re-running the scenario in a Codex session and confirming that descriptions stay prose-only and unprompted `sudo` prefixes disappear—needs a human operator and has not been done.

Closes #172


tools/prompts.py
```diff
@@ -24,10 +24,13 @@ def work_session_workflow(intent: str, servers: str = '') -> str:
    - Required arguments: `workspace`, `scopes` (list), `servers` (list of UUIDs),
      `expires_at` (ISO 8601), and `description` — fill them all or the call is rejected.
    - Pass the goal above as the session `description` (the API has no `intent` field).
+     The description is prose for the human who approves the session — never a list of
+     commands. Nothing in it is executed; commands run via `execute_command` in step 3.
```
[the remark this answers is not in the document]
jisung-02: Fixed in 53c33b8. The file was uniformly spaced (18 of 18 em dashes), so normalizing only the added lines would have left it mixed — every em dash in `tools/prompts.py` now drops its surrounding spaces, matching the CLAUDE.md rule.

tools/prompts.py
```diff
@@ -24,10 +24,13 @@ def work_session_workflow(intent: str, servers: str = '') -> str:
...
    - Valid scopes are `command`, `webftp`, `tunnel`, and `sudo`. As an agent (MCP channel)
      you can request `command`/`webftp`/`tunnel` directly; `sudo` is available but every
-     sudo invocation routes to human approval (HITL). Interactive `websh`/`editor` access
-     requires human presence (MFA) and is NOT available to you.
+     sudo invocation routes to human approval (HITL), so request it only when the work
+     genuinely requires root. Interactive `websh`/`editor` access requires human presence
+     (MFA) and is NOT available to you.
```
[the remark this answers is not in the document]
jisung-02: Fixed in 53c33b8. Both spots now carry the same qualifier the tool descriptions use: a sudo invocation routes to human approval *unless a Work Session sudo policy already covers the command*, and some are denied outright with no request anyone can approve. Step 3 of `guarded_execution` names `SUDO_RISK_DENIED` explicitly and tells the agent to read the returned category before it starts waiting.

tools/command_tools.py
```diff
@@ -274,7 +277,7 @@ async def list_commands(
...
+    description='Run a shell command on a server and wait for the result (up to 5 minutes by default). Returns stdout, stderr, and exit code in a single call. Requires ACL permission. Do not prefix the command with sudo by default: unless a Work Session sudo policy already covers the command, a sudo invocation routes to human-in-the-loop approval and blocks until a human acts. Use sudo only when the command genuinely requires root and the Work Session carries the "sudo" scope. The timeout resets when the command is actively running. Supports dependency chains (run_after), scheduled execution (scheduled_at), and stdin data. Pass work_session_id to link this command to a Work Session for audit—the server enforces this for MCP OAuth and browser-based auth. When to use: the recommended way to run a command on a server. Related: execute_command_multi_server (run on multiple servers), list_commands (browse history), work_session_create (create a Work Session). Note: Default timeout is 300 seconds (5 minutes).',
```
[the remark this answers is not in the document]
jisung-02: Fixed in 53c33b8. The MFA-bypass case was already covered by "unless a Work Session sudo policy already covers the command"; the hard-denial case was not. Both descriptions now read "either routes to human-in-the-loop approval and blocks until a human acts, or is denied outright with no request anyone can approve. Read the returned category before waiting."

tools/command_tools.py
```diff
@@ -460,7 +463,7 @@ async def execute_command(
...
+    description='Run the same shell command on multiple servers simultaneously or sequentially. Returns per-server results with success/failure status. Requires ACL permission. Do not prefix the command with sudo by default: unless a Work Session sudo policy already covers the command, a sudo invocation either routes to human-in-the-loop approval and blocks until a human acts—once per server—or is denied outright with no request anyone can approve. Read the returned category before waiting. Use sudo only when the command genuinely requires root and the Work Session carries the "sudo" scope. Pass work_session_id to link commands to a Work Session for audit—the server enforces this for MCP OAuth and browser-based auth. When to use: batch operations like deploying configs, checking status, or running diagnostics across a fleet. Related: execute_command (single server), work_session_create (create a Work Session). Note: Set parallel=false for sequential execution. This submits commands without waiting for results — use list_commands to check status.',
```
junho226: Two sentences in this description now disagree. This one says a sudo invocation `blocks until a human acts—once per server`; the last sentence says `This submits commands without waiting for results — use list_commands to check status`.

Both are true of different subjects—the tool call returns immediately, the *command* is what stalls on the host—but an agent reading top to bottom takes away "this call will hang." `execute_command` genuinely blocks (it waits up to 300s), so the phrase does not survive the copy over to the submit-only tool.

Suggestion: `...either routes to human-in-the-loop approval and stalls that server's command until a human acts—once per server—or is denied outright...`

Separate nit on the same line: `results — use list_commands` keeps a spaced em dash while this PR normalizes them to the unspaced form in `tools/prompts.py` per CLAUDE.md. Pre-existing text, but the line was rewritten anyway.

junho226 (dismissed):
## Review

**Approve.** Description text only—I diffed it and confirmed no control flow, request shape, or parameter changed. The steering is well built, and more importantly it is accurate about what the platform actually does.

### Does it work as steering

The `work_session_create` / `work_session_update` wording is the strong part. It states the rule positively first (prose for the human who approves), then negates, then redirects to the correct channel (commands run via `execute_command`). Giving the agent somewhere to put the command is what stops the misuse; a bare "do not paste commands here" would not.

The sudo passage leads with the negation but immediately supplies the cost (blocks until a human acts, or is denied outright) and closes with the positive rule (use sudo only when the command genuinely requires root and the session carries the `sudo` scope). That is the shape that works.

Consistency across channels is clean: the four tool descriptions, `_SUDO_DENIAL_HINTS['SUDO_INTENT_DEVIATION']`, `_NEXT_ACTION_BY_CATEGORY`, and both prompts carry the same rule, and `grep -rni 're-declare'` returns nothing—no channel left behind, which is the usual failure mode for this kind of change. `execute_command` and `execute_command_multi_server` are also the only two tools taking an executable command (`security_tools` and `create_sudo_policy` take ACL/policy *patterns*), so there is no permissive sibling for an agent to route around.

### Steering vs enforcement

Worth stating plainly since the PR body does not: this is not a control and does not claim to be. A tool description is advisory—a model can ignore it, and a prompt-injected model will.

But the check it describes does exist, downstream. `_SUDO_DENIAL_HINTS` is documented as kept in sync with alpacon-server `utils/error_codes.py`, and the categories arrive from `alpacon_approval.c` / `pam_alpamon.c` on the host. Sudo is enforced by the Work Session sudo policy, the risk assessment, and PAM—not here. So these descriptions now *describe* an enforced control rather than promise an unenforced one. Nothing in the new text is a promise the code fails to keep, and the qualifier work in commit 53c33b8 is what keeps it that way.

The other claim—that `description` is not an execution channel—needs no enforcement: the field is passed to the API as a string and is never executed. The failure was agents believing otherwise, so steering is the correct and only fix. The second-order value is real too: a description full of command text degrades the audit record the approver reads.

I agree with the decision recorded in the PR body not to add a heuristic guard that rejects command-shaped descriptions.

### Findings

1. (inline) `execute_command_multi_server`: "blocks until a human acts" contradicts "This submits commands without waiting for results" later in the same description.
2. Follow-up, not this PR: **no tool in this server carries parameter-level descriptions.** I dumped the served schemas and every property is bare—`command` is `{"title": "Command", "type": "string"}`, `description` likewise. All guidance lives in the tool description. `anthropic/alwaysLoad: True` covers `execute_command`, but `work_session_create` has no such guarantee and its rule sits mid-paragraph in a long description. Moving the per-field rules onto the fields (`Annotated[str, Field(description=...)]`) would put them where clients render them at argument-fill time. Repo-wide change, correctly out of scope here.
3. Minor: the new tests pin exact literal phrases (`'Do not prefix the command with sudo by default'`). That is a deliberate regression guard, but it also pins the current negative framing—a later rewording toward positive-only phrasing fails the test rather than passing it. Fine as-is, just know they are text locks, not semantic ones. Good call asserting against `mcp.list_tools()` rather than the source strings.

Also noting the `tools/prompts.py` em-dash normalization reaches `security_audit` and `incident_response`, hunks unrelated to sudo or descriptions. The commit message justifies it and I would not undo it, but the repo keeps spaced em dashes in about ten other files (`tools/workspace_tools.py`, `tools/resources.py`, `utils/decorators.py`, ...), so the rule is not uniform anyway.

### Checks run

| Check | Result |
|---|---|
| `uv run pytest -q` | `1092 passed, 7 warnings in 7.13s` |
| `uvx ruff check tools/ utils/ tests/` | `All checks passed!` |
| `uvx ruff format --check tools/ utils/ tests/` | `77 files already formatted` |
| `uv run mypy tools/ utils/` | `Found 8 errors in 1 file (checked 34 source files)`—all in `utils/oauth.py`, **pre-existing**: identical count on `origin/main`, and no mypy job in CI |
| Served-schema dump via `mcp.list_tools()` | read all four descriptions as an agent receives them; new text present, no property-level description anywhere |
| New test classes in isolation | both pass when their file runs alone, so the module-level registration imports hold |
| CI | all 6 checks green |

**Not run:** the behavioral verification the PR body already flags—re-running the scenario in a live Codex session to confirm descriptions stay prose-only and unprompted `sudo` prefixes actually drop. That needs a human operator and a live workspace. Since the change is measured in agent behavior, that is the check that would actually confirm it works; everything above only confirms the text is correct, consistent, and served.

tools/command_tools.py
jisung-02: Fixed in db13c91, both parts.

Took the suggested wording: `...routes to human-in-the-loop approval and stalls that server's command until a human acts—once per server—or is denied outright...`. You are right that the phrase did not survive the copy: `execute_command` blocks for up to 300s and "blocks" is literal there, while this tool returns as soon as the commands are submitted. Left `execute_command` unchanged for that reason.

Also unspaced the em dash on the same line: `without waiting for results—use list_commands to check status`.

On the two out-of-scope items: the parameter-level descriptions are now #175, with the `work_session_create` / `execute_command` arguments as the first slice and a note to confirm `Annotated` metadata survives `@mcp_tool_handler` into the served schema before writing any of it. The text-lock point on the new tests is fair and I am leaving them as they are; a reword should have to be deliberate.

`uv run pytest -q`: `1092 passed, 7 warnings in 6.49s`. `uvx ruff check tools/ utils/ tests/`: `All checks passed!`

tools/command_tools.py
```diff
@@ -274,7 +277,7 @@ async def list_commands(
...
+    description='Run a shell command on a server and wait for the result (up to 5 minutes by default). Returns stdout, stderr, and exit code in a single call. Requires ACL permission. Do not prefix the command with sudo by default: unless a Work Session sudo policy already covers the command, a sudo invocation either routes to human-in-the-loop approval and blocks until a human acts, or is denied outright with no request anyone can approve. Read the returned category before waiting. Use sudo only when the command genuinely requires root and the Work Session carries the "sudo" scope. The timeout resets when the command is actively running. Supports dependency chains (run_after), scheduled execution (scheduled_at), and stdin data. Pass work_session_id to link this command to a Work Session for audit—the server enforces this for MCP OAuth and browser-based auth. When to use: the recommended way to run a command on a server. Related: execute_command_multi_server (run on multiple servers), list_commands (browse history), work_session_create (create a Work Session). Note: Default timeout is 300 seconds (5 minutes).',
```
junho226: `Read the returned category before waiting.` is the load-bearing half of this addition, but there is no `category` field for a schema-only client to read.

`_attach_sudo_denial` below (line 147) always sets `sudo_hint`, and sets the structured object only for codes in `_SUDO_HUMAN_APPROVAL_CODES` — nested as `sudo_denial.category`, never top-level `category`. `SUDO_RISK_DENIED` is deliberately not in that set, so on the hard denial this sentence exists for, the response carries `sudo_hint` and no `sudo_denial` key at all. `tests/test_command_tools.py:380-381` locks exactly that shape.

`tools/prompts.py` step 4 names the fields correctly, but prompts being opt-in is this PR's own reason for moving the rule here, so the description cannot lean on the prompt.

Still description-only to fix — something like: `Check sudo_denial.category before waiting; a sudo_hint with no sudo_denial is a hard denial that creates no request, so do not wait on it.`

tools/command_tools.py
```diff
@@ -460,7 +463,7 @@ async def execute_command(
...
+    description='Run the same shell command on multiple servers simultaneously or sequentially. Returns per-server results with success/failure status. Requires ACL permission. Do not prefix the command with sudo by default: unless a Work Session sudo policy already covers the command, a sudo invocation either routes to human-in-the-loop approval and stalls that server\'s command until a human acts—once per server—or is denied outright with no request anyone can approve. Read the returned category before waiting. Use sudo only when the command genuinely requires root and the Work Session carries the "sudo" scope. Pass work_session_id to link commands to a Work Session for audit—the server enforces this for MCP OAuth and browser-based auth. When to use: batch operations like deploying configs, checking status, or running diagnostics across a fleet. Related: execute_command (single server), work_session_create (create a Work Session). Note: Set parallel=false for sequential execution. This submits commands without waiting for results—use list_commands to check status.',
```
junho226: Same as on `execute_command` above: `Read the returned category before waiting.` points at a field that is absent on the hard-denial path. Whatever wording lands there should land here too, so the two tools keep saying the same thing.

The `stalls that server's command until a human acts—once per server` phrasing is right, and `db13c91` was the correct call — "blocks" would have contradicted the closing `This submits commands without waiting for results` on the same line.

tests/test_command_tools.py
```diff
@@ -1070,3 +1071,19 @@ async def test_work_session_required_translated_in_parallel(
...
+
+    @pytest.mark.asyncio
+    @pytest.mark.parametrize(
+        'tool_name', ['execute_command', 'execute_command_multi_server']
+    )
+    async def test_description_states_the_sudo_cost(self, tool_name):
+        descriptions = {t.name: t.description for t in await mcp.list_tools()}
+        text = descriptions[tool_name]
+        assert 'Do not prefix the command with sudo by default' in text
+        assert 'human-in-the-loop' in text
```
junho226, line 1089: This assertion is not load-bearing: `human-in-the-loop` appears in any description containing the phrase, including one whose surrounding sentence claims the opposite. It also leaves the two clauses the Copilot round produced — the policy carve-out and the hard-denial path — completely unpinned, so they can regress without failing anything.

Asserting the substance instead would lock what the last two commits were written to add:

```python
assert 'sudo policy already covers' in text
assert 'denied outright' in text
```

Asserting against `mcp.list_tools()` rather than the source constant is the right level, though — that part catches a description that fails to register.

junho226 (commented):
## What this PR does

Description text only. It moves the "do not prefix `sudo` by habit" rule out of `tools/prompts.py` (opt-in, so a schema-only client never sees it) into the `execute_command` / `execute_command_multi_server` tool descriptions, and states in `work_session_create` / `work_session_update` that `description` is prose for the approver and that nothing in it is executed. The `SUDO_INTENT_DEVIATION` correction lands on both channels that carry it — the free-text hint in `tools/command_tools.py` and the structured `next_action` in `utils/common.py`. No control flow, no request shape, no server-side behavior.

**Verdict: comment.** The targeting is right and the reasoning in the PR body survives checking — I verified the sudo-policy carve-out and the hard-denial claim against the code rather than taking them on trust. One finding is worth fixing before merge, because it sits inside the exact sentence the last two commits were written to add.

## Findings

### 1. (medium) "Read the returned category before waiting" names a field a schema-only client cannot find

Both new descriptions close the sudo rule with `Read the returned category before waiting.` That instruction is the load-bearing half of commit `53c33b8` — it is what keeps an agent from waiting forever on a hard denial. But there is no `category` in the response the sentence points at:

- `_attach_sudo_denial` (`tools/command_tools.py:147-165`) always sets free-text `sudo_hint`, and sets the structured object **only** when the code is in `_SUDO_HUMAN_APPROVAL_CODES` — and then nested, as `sudo_denial.category`, never as a top-level `category`.
- `SUDO_RISK_DENIED` is deliberately **not** in that set. So on a hard denial the response carries `sudo_hint` and no `sudo_denial` key at all. `tests/test_command_tools.py:380-381` locks exactly that shape (`assert 'risk' in entry['sudo_hint']` / `assert 'sudo_denial' not in entry`).

So in the one case the sentence was added for, the field it tells the agent to read is absent. `tools/prompts.py` step 4 states the shape correctly ("the field is `code` on an `error` result and `category` on a `pending_approval` one") — but prompts being opt-in is this PR's own argument for moving the rule into the description, so the description cannot lean on the prompt to fill the gap.

The fix is one clause, still description-only. Something like: `Check sudo_denial.category before waiting; a sudo_hint with no sudo_denial is a hard denial that creates no request, so do not wait on it.` That makes the same point actionable for a client that only ever sees the schema.

The routing itself is correct as designed — I am not asking for a code change, only for the description to name the shape the code actually returns.

### 2. (low) `docs/api-reference.md` was not updated, so the rule now lives on three of four surfaces

This repo's `CLAUDE.md` calls that file "the catalog this file deliberately does not duplicate. Read it before calling an unfamiliar tool or adding one." It still documents the parameter as `` `command` (string): Command to execute `` (`docs/api-reference.md:343`), and the `work_session_create` entry (`:624`) still describes `description` with no note that it is inert. Two sentences close it, and it keeps a human operator or a docs-reading agent on the same rule as a schema-reading one.

While you are on line 624: it lists the agent-requestable scopes as `command`, `webftp`, `tunnel`, while `work_session_workflow` says `sudo` is directly requestable by an agent ("`sudo` is available but …"). That divergence predates this PR, but it is on the line you would be editing and it is about the same thing.

### 3. (low) the second assertion in the new test is not load-bearing

`assert 'human-in-the-loop' in text` (`tests/test_command_tools.py:1089`) passes on any description containing that phrase, including one whose surrounding sentence says the opposite. It also does not cover the two clauses the last two commits exist to add — the policy carve-out and the hard-denial path. Asserting `'sudo policy already covers'` and `'denied outright'` would actually pin the wording that came out of the Copilot round; as written, that wording is unlocked and can regress silently.

### 4. (nit) `execute_command`'s always-loaded description grew 724 → 1131 chars (+56%)

It carries `anthropic/alwaysLoad: True`, so this is a permanent per-session context cost, not a per-call one. I think it is the right trade for the failure it prevents — recording it so the next addition to this particular string gets weighed rather than appended.

### 5. (nit) one spaced em dash survives in the file you normalized

`utils/common.py:95` still reads `'decision — there is nothing for you to retry or resubmit.'`, in the same dict this PR edits, while `tools/prompts.py` was normalized per `CLAUDE.md`. (`tools/command_tools.py:434` too, in a comment.) Outside the diff, so entirely optional — flagging only because the em-dash pass was explicit in commit `53c33b8`.

## On the questions this review was asked

**Does the wording remove the ambiguity, or is it a "do not" that models ignore?** It is better than a bare prohibition. It states the cost ("blocks until a human acts"), the carve-out ("unless a Work Session sudo policy already covers the command"), and the positive rule ("Use sudo only when the command genuinely requires root and the Work Session carries the `sudo` scope"). That is the shape that actually moves a model. The one thing still missing is the fallback at the moment of choice: what to do when the agent is *unsure* whether root is needed. "Run it without `sudo` first and escalate only after it fails on permissions" is the instruction that resolves that, and it costs one clause. Optional, but it is the highest-leverage sentence still not in there.

**Is it enforced anywhere, and is description-only acceptable for security-adjacent behavior?** Description-only, and here that is the right boundary. Enforcement lives server-side in alpacon-server's approval gate and risk assessment; a client-side reject in this MCP server would be advisory anyway and would break legitimate root work. Worth separating from the guard the PR body rejects: that one was a heuristic on whether a *description* looks like a command block, which is genuinely lossy. A `command.strip().startswith('sudo ')` check is not that — it is deterministic, and `_attach_sudo_denial` already demonstrates the non-blocking pattern (annotate the response, do not reject the call). A pre-flight note when a `sudo`-prefixed command meets a session with no `sudo` scope would put the nudge at the moment it matters instead of in a description the model read a hundred turns ago. Follow-up issue alongside #175, not a change for this PR.

**Every tool that takes a command, none missed?** Verified. `execute_command` and `execute_command_multi_server` are the only execution surfaces in this server — the only other `command: str` parameters are `security_tools.create_command_acl` / `update_command_acl`, which take ACL *patterns* rather than executing anything and are correctly untouched. There is no websh or terminal tool here. The divergence between the two descriptions ("blocks until a human acts" vs "stalls that server's command until a human acts—once per server") is justified and `db13c91` fixed it in the right direction: `execute_command_multi_server` submits without waiting, so "blocks" would have read as the call hanging.

**Examples consistent with the new rule?** Yes. `work_session_create`'s in-description example ("Restart nginx on prod-web-1 to clear 502s; no package installs") is prose and models the rule rather than contradicting it. No `sudo`-prefixed `execute_command` example anywhere in the docs cuts against it. The old `re-declare` wording is fully gone — `grep -rn 're-declare\|Re-declar' tools/ utils/ tests/ docs/ README.md` returns nothing.

**Schema tests updated?** Yes, and at the right level: both new classes assert against what `mcp.list_tools()` actually returns rather than the source constants, so a description that fails to register is caught. See finding 3 on how tightly they pin it.

## Checks run

Run in a worktree at `db13c91`, the PR head. Commands are the ones `.github/workflows/test.yml` runs.

| Command | Result |
|---|---|
| `uv run --no-sync pytest tests/ -q --tb=short` | `1092 passed, 7 warnings in 6.16s` |
| `uv run --no-sync ruff check .` | `All checks passed!` |
| `uv run --no-sync ruff format --check .` | `82 files already formatted` |
| `uv run --no-sync mypy --ignore-missing-imports --no-strict-optional .` | `Found 8 errors in 1 file (checked 82 source files)` |

The 8 mypy errors are all in `utils/oauth.py:422` and `:437` (`set_cookie` / `delete_cookie` called with `**dict[str, object]`), a file this PR does not touch. CI marks that step `continue-on-error: true`, so it does not gate.

**Not run:** CI's Python 3.13 and 3.14 matrix legs — the local venv is 3.12.9, so I exercised only the 3.12 leg. The PR changes string literals, so a version-specific failure is unlikely, but I did not verify it. I also did not run the coverage gate (`--cov-fail-under=60`) and cannot confirm the coverage number CI computes.

The behavioral verification the PR body already flags as outstanding — re-running the Codex scenario and confirming descriptions stay prose-only and unprompted `sudo` prefixes disappear — is the one that decides whether this actually worked, and it needs a human operator. Nothing here substitutes for it.

tools/command_tools.py
jisung-02: Fixed in `a7db50a`. Both descriptions now close with your wording: `Check sudo_denial.category before waiting; a sudo_hint with no sudo_denial is a hard denial that creates no request, so do not wait on it.`

The finding holds as written — I re-read `_attach_sudo_denial` (`tools/command_tools.py:147-165`) and `_SUDO_HUMAN_APPROVAL_CODES` rather than taking it on trust. `sudo_hint` is unconditional, the structured object is gated on the code being in that set, and it lands nested. There is no top-level `category` anywhere on the response, so the sentence was pointing at nothing precisely on the `SUDO_RISK_DENIED` path it existed to cover. `tests/test_command_tools.py:380-381` does lock that shape, which is what makes it a description bug rather than a code one.

Your point about not leaning on `tools/prompts.py` is the one that decided the fix. Step 4 is correct there, but "prompts are opt-in" is this PR's whole argument for moving the rule into the description, so borrowing the prompt's accuracy would have been circular.

tools/command_tools.py
jisung-02: Same fix landed here in `a7db50a`, character-for-character identical to the clause on `execute_command`. The two tools should not diverge on how to read a denial, and the new test is parametrized over both names, so a future edit to one and not the other fails.

Thanks for confirming the `stalls that server's command until a human acts—once per server` direction. That was the reading I was least sure of when I wrote `db13c91` — the closing `This submits commands without waiting for results` on the same line is what made "blocks" untenable.

tests/test_command_tools.py
jisung-02: Fixed in `a7db50a`. Took both assertions you named and added the field name from finding 1, so the test now pins all three clauses the review rounds produced:

```python
assert 'sudo policy already covers' in text
assert 'denied outright' in text
assert 'sudo_denial.category' in text
```

Kept `human-in-the-loop` rather than replacing it. You are right that it is not load-bearing on its own, but it is the phrase the routing behavior is named by, and it costs nothing beside three assertions that now carry the weight.

The wording is deliberately locked now. Anyone rephrasing these descriptions has to update the test in the same commit, which is the point.

jisung-02:
The three findings without an inline thread are handled too, at `af51ead`.

**Finding 2 (docs/api-reference.md).** `e6880e3`. The `command` parameter on both execution tools now carries the sudo rule and the `sudo_denial.category` shape, `work_session_create` gets a paragraph saying `description` is prose for the approver and that nothing in it is executed, and `work_session_update` points at that same rule. I also took the divergence you flagged on the line: `sudo` is now listed among the agent-requestable scopes, with the caveat that a sudo invocation no policy covers routes to human approval. That leaves the catalog saying what `work_session_workflow` has been saying.

**Finding 4 (description growth).** Recording it rather than acting on it, as you framed it. Worth noting the direction moved against me: finding 1 pushed `execute_command` from 1131 to 1273 characters, another +13% on an always-loaded string. I still think the trade holds — an agent waiting forever on a denial that creates no request is the expensive failure — but the next addition to this string should have to argue for itself.

Which is why I did not take the optional clause from your closing section. "Run it without `sudo` first and escalate only after it fails on permissions" is the highest-leverage sentence still missing, I agree, but adding it in the same round as finding 1 means two growths on the always-loaded description in one PR with no measurement between them. Better as its own change where it can be weighed alone.

**Finding 5 (em dashes).** `af51ead`. `utils/common.py:95` and the comment at `tools/command_tools.py:434`, both to the unspaced form.

The pre-flight `command.strip().startswith('sudo ')` idea is the one I want to keep. Your separation is the right one — it is deterministic where the description-shaped guard was a heuristic, and `_attach_sudo_denial` already demonstrates annotating a response instead of rejecting a call. Filed as #177, alongside #175.

Verification at `af51ead`: `uv run pytest -q` gives `1092 passed, 7 warnings in 6.17s`, `uvx ruff check tools/ utils/ tests/` gives `All checks passed!`, `uvx ruff format --check tools/ utils/ tests/` gives `77 files already formatted`. Same caveat as yours on the 3.13/3.14 matrix legs and the coverage gate — CI covers those, I did not run them locally. The behavioral verification in a live Codex session is still outstanding and still needs a human operator.

tools/command_tools.py
```diff
@@ -460,7 +463,7 @@ async def execute_command(
...
+    description='Run the same shell command on multiple servers simultaneously or sequentially. Returns per-server results with success/failure status. Requires ACL permission. Do not prefix the command with sudo by default: unless a Work Session sudo policy already covers the command, a sudo invocation either routes to human-in-the-loop approval and stalls that server\'s command until a human acts—once per server—or is denied outright with no request anyone can approve. Check sudo_denial.category before waiting; a sudo_hint with no sudo_denial is a hard denial that creates no request, so do not wait on it. Use sudo only when the command genuinely requires root and the Work Session carries the "sudo" scope. Pass work_session_id to link commands to a Work Session for audit—the server enforces this for MCP OAuth and browser-based auth. When to use: batch operations like deploying configs, checking status, or running diagnostics across a fleet. Related: execute_command (single server), work_session_create (create a Work Session). Note: Set parallel=false for sequential execution. This submits commands without waiting for results—use list_commands to check status.',
```
junho226: **(low) `sudo_denial.category` is not a field this tool's response carries.**

The sentence is character-for-character the one on `execute_command`, and that is what makes it slightly off here. `_attach_sudo_denial` has exactly two call sites — `tools/command_tools.py:268` (`list_commands`, per entry) and `:398` (`execute_command`) — and neither is on this path. `execute_command_multi_server` puts `deploy_shell_results[sid] = {'status': 'success', 'data': <submit response>}` straight from `_submit_command`, and `_sudo_denial` reads `result['result']`, which at submit time holds no output yet. So `sudo_hint` and `sudo_denial` cannot appear on this response by construction, not by oversight.

The information is reachable, which is why this is a nit rather than a repeat of finding 1: `list_commands` already states it correctly at `:228` ("An entry whose output is a sudo denial also carries sudo_hint and, when a human can clear it, a structured sudo_denial block"), and this description does close with `use list_commands to check status`. But that pointer sits ~130 characters after the sentence telling the agent to check the field, and a schema-only client reading top to bottom will look for it on the value in hand — the same failure mode as finding 1, one hop removed.

Wording-only, e.g.:

> ...or is denied outright with no request anyone can approve. The denial surfaces on that server's `list_commands` entry: check `sudo_denial.category` there before waiting; a `sudo_hint` with no `sudo_denial` is a hard denial that creates no request, so do not wait on it.

That keeps the `sudo_denial.category` substring, so the new parametrized assertion still passes unchanged.

junho226 (commented):
Second pass, `db13c91` → `af51ead` (3 commits). All five findings from the previous review are resolved or consciously recorded, and finding 1 was fixed where I hoped it would be — in the wording, not in the response shape. One new low-severity point, introduced by the copy that fixed finding 1.

## Previous findings

**1 (medium) — fixed.** Both descriptions now close the sudo rule with `Check sudo_denial.category before waiting; a sudo_hint with no sudo_denial is a hard denial that creates no request, so do not wait on it.` That is accurate for `execute_command`: `_attach_sudo_denial(response, result)` (`tools/command_tools.py:398`) sets `sudo_hint` unconditionally and `sudo_denial` only for codes in `_SUDO_HUMAN_APPROVAL_CODES`, so "hint present, block absent" is exactly the hard-denial signature the sentence now describes.

The fix stayed description-only, which is what makes it cheap to trust. `_SUDO_HUMAN_APPROVAL_CODES` is byte-identical, `SUDO_RISK_DENIED` is still excluded, and the tests that lock the shape (`tests/test_command_tools.py:380-381`) are untouched. `git diff db13c91..af51ead -- tools/command_tools.py` is two description strings plus one comment — no behavior moved, and nothing else in the tree keyed off that exclusion, so there was nothing to keep coherent.

**2 (low) — fixed** in `e6880e3`. `docs/api-reference.md` now carries the rule on the `command` parameter of both execution tools (`:343`, `:370`), the `description`-is-prose-for-the-approver paragraph on `work_session_create` (`:624-626`), and the pointer to it from `work_session_update` (`:634`). Listing `sudo` among the agent-requestable scopes also closes the divergence with `work_session_workflow` I flagged on that line. The rule now lives on all four surfaces.

**3 (low) — fixed** in `a7db50a`. The three added assertions pin the substance rather than a phrase that passes trivially:

```python
assert 'sudo policy already covers' in text
assert 'denied outright' in text
assert 'sudo_denial.category' in text
```

Keeping `human-in-the-loop` alongside them is fine — it was only a problem while it was carrying the weight alone. Parametrized over both tool names, so the two descriptions can no longer drift apart silently.

**4 (nit) — measured, not acted on**, as framed. Re-measured at `af51ead` by parsing the `description=` string literals with `ast` at each revision; the decorator passes the string through unchanged (`utils/decorators.py:545`), so the literal is what ships:

| revision | `execute_command` | `execute_command_multi_server` |
|---|---|---|
| merge-base `e8eea1b` | 703 | 623 |
| `db13c91` | 1109 | 1064 |
| `af51ead` | **1205** | **1160** |

Finding 1 cost +96 chars (+8.7%) on the always-loaded string; the PR as a whole takes it +71%. My absolute numbers sit below the ones quoted in the thread — different counting basis — but the deltas agree in direction and rough size, and the trend is the part worth tracking. Agreed on holding the "run it without `sudo` first" clause for its own change rather than stacking a second growth here with no measurement between them.

**5 (nit) — fixed** in `af51ead`, both spots. The two spaced em dashes still present in files this PR touches (`utils/common.py:37`, `docs/api-reference.md:836`) sit on lines the PR never rewrote; leaving them alone is the right call.

## New

**(low) the multi-server copy names fields that tool's own response never carries** — inline on `tools/command_tools.py:466`. Wording-only, not blocking, and it can equally ride in the follow-up.

## Checks run

In a worktree at `af51ead`, Python 3.12, all via `uv run --no-sync`:

- `pytest tests/ -q` → `1092 passed, 7 warnings in 6.69s` — unchanged from the previous head.
- The full CI invocation including the coverage gate, `pytest tests/ --tb=short --cov=. --cov-report=term-missing --cov-fail-under=60` → `Required test coverage of 60% reached. Total coverage: 94.85%`, `1092 passed`.
- `ruff check .` → `All checks passed!`
- `ruff format --check .` → `82 files already formatted`
- `mypy --ignore-missing-imports --no-strict-optional .` → `Found 8 errors in 1 file (checked 82 source files)`, all at `utils/oauth.py:422`/`:437`. Unchanged from the previous head, in a file this PR does not touch, and the CI step is `continue-on-error: true`.

Not run: the 3.13 and 3.14 matrix legs — this worktree only has 3.12, so CI remains the only signal there. And the behavioral check that a live Codex session actually stops emitting `sudo` prefixes is still outstanding and still needs a human operator; nothing in this suite substitutes for it.

Good to merge from my side.

tools/command_tools.py
jisung-02: Fixed in `28aaec0`, with your wording:

> The denial surfaces on that server's list_commands entry: check sudo_denial.category there before waiting; a sudo_hint with no sudo_denial is a hard denial that creates no request, so do not wait on it.

Confirmed the mechanism before editing rather than trusting the copy. `_attach_sudo_denial` has exactly the two call sites you name — `:268` and `:398` — and `execute_command_multi_server` fills `deploy_results[sid]` at `:531` and `:548` directly from `_submit_command`, with nothing in between. The fields cannot appear there, so the sentence was unreachable rather than merely imprecise.

Your framing of it as one hop removed from finding 1 is the part I took. The description did already end with `use list_commands to check status`, and I had treated that as sufficient — it is not, when the sentence 130 characters earlier tells the agent to check a field and the agent has a response in hand.

`execute_command` keeps the original sentence unchanged; `_attach_sudo_denial(response, result)` at `:398` puts the fields on that response directly, so the extra hop would be wrong there. The two descriptions now differ on purpose in one more place, same as the `blocks` / `stalls that server's command` split. The parametrized assertion still passes on both — `sudo_denial.category` survives in each.

`1092 passed, 7 warnings in 7.32s`, `All checks passed!`, `20 files already formatted`. Thanks for the coverage-gate number on `af51ead`; I had not run that leg and have added it to the PR body.

junho226 (approved):
Third pass, reviewing `af51ead..28aaec0` only — one commit, one line, description text.

**The open point is resolved.** `28aaec0` rewrites the `execute_command_multi_server` clause to `The denial surfaces on that server's list_commands entry: check sudo_denial.category there before waiting; ...`. The fix is in the wording, not the response shape, which is the right half to change: the fields genuinely cannot appear on this response, so making them appear would have been the wrong repair. I re-derived the mechanism rather than reading it off the commit message — `_attach_sudo_denial` still has exactly two call sites (`tools/command_tools.py:268` in `list_commands`, `:398` in `execute_command`), and `execute_command_multi_server` fills `deploy_results[sid]` at `:531`/`:548` straight from `_submit_command` with nothing in between. Nothing else in the tree writes `sudo_hint` or `sudo_denial`. The new pointer also lands somewhere that already documents the shape: `list_commands` states it at `:228`, so the two descriptions agree.

Because no code changed, there was no old response shape for tests to pin, and `_SUDO_HUMAN_APPROVAL_CODES` / `SUDO_RISK_DENIED` handling is untouched — `SUDO_RISK_DENIED` remains outside the frozenset, and `tests/test_command_tools.py:380-381` still locks the hint-without-block shape that the sentence describes. The parametrized `assert 'sudo_denial.category' in text` (`:1094`) passes unchanged on both tools, as intended. Keeping `execute_command`'s sentence in the direct form is correct, not an inconsistency: `_attach_sudo_denial(response, result)` puts the fields on that response, so routing it through `list_commands` there would have been a false hop.

One note in the PR's favor on the description-growth concern recorded under finding 4: this change is +65 characters (1161 → 1226) on `execute_command_multi_server`, and `anthropic/alwaysLoad` appears exactly once in the file, on `execute_command` at `:283`. So the growth landed on a description that is not always loaded. Nothing to act on.

No new findings. This looks ready to merge.

### Checks run

At `28aaec0`, from the repo root, matching `.github/workflows/test.yml`:

- `uv run --no-sync pytest tests/ -q` → `1092 passed, 7 warnings in 7.38s` (unchanged from the `af51ead` baseline)
- `uv run --no-sync ruff check .` → `All checks passed!`
- `uv run --no-sync ruff format --check .` → `82 files already formatted`
- `uv run --no-sync mypy --ignore-missing-imports --no-strict-optional .` → `Found 8 errors in 1 file (checked 82 source files)`, all in `utils/oauth.py:422`/`:437`, untouched by this PR and identical to the baseline. CI marks that step `continue-on-error: true`.

Not run locally: the 3.13/3.14 matrix legs and the `--cov-fail-under=60` coverage gate. CI covers both. The live-Codex behavioral verification the author flagged as outstanding still needs a human operator — a description-only change cannot be confirmed by this suite.

Issue #160, opened 2026-08-05
Follow-up to #157: per-flow nonce cookie name, stage-2 comment, rollout discontinuity note


jisung-02:
Follow-up to #157 (OAuth state browser binding), collecting the review items from @junho226 that we agreed to handle after merge.

## 1. Fixed nonce cookie name breaks concurrent OAuth flows in the same browser

The nonce cookie name (`__Host-alpacon_oauth_nonce`) is fixed, so when two flows overlap in the same browser, the second `authorize` overwrites the first flow's cookie, and the first flow's callback then answers 400.

Reproduction from the review (verified by the reviewer):

```
flow1 authorize -> flow2 authorize (same browser)
flow1 cookie == flow2 cookie ?  False   (overwritten)
flow1 callback                  ->  400 invalid_request
```

The failure recovers on retry, and a user double-clicking through the flow is harmless (the first flow is abandoned anyway). The real case is **two different MCP clients authenticating through the same browser at the same time**: one side fails for no visible reason, and the client-facing message is `Invalid or expired state parameter`, which points suspicion at the state when the state is fine and the cookie was overwritten. (The response message itself is correct as is—no reason to tell an attacker which check fired—and the server log already distinguishes the case as `state not bound to this browser`, so a debugging path exists.)

Proposed fix (from the review): add a per-flow suffix to the cookie name and carry that suffix inside the signed state, so the callback knows which cookie to look at.

Key phrasing for whoever picks this up: it is not "concurrent flows break", it is "the second authorize overwrites the first flow's cookie, so the first callback gets 400".

## 2. Make the stage-2 cookie access self-explanatory (`utils/oauth.py:1013`)

The MFA stage-2 re-set subscripts `request.cookies` directly instead of using `.get()`. This is safe today: `stage` only comes from `state_data`, which is read behind the nonce gate, so any request reaching stage 2 has already passed `_nonce_cookie_matches`, which implies a non-empty cookie. But that coupling is invisible at the line itself, and a later change that moves the gate or hoists the stage-2 branch could silently turn it into a 500. Switching to `.get()` is not the answer (it would set an empty cookie); a one-line comment stating "the gate has passed, so the cookie must exist" is enough.

## 3. Document the rollout discontinuity

[...]

PR #180, opened 2026-08-14
fix(oauth): type the shared nonce cookie attributes as a TypedDict

### LINKED ISSUE #167: fix(oauth): type the shared nonce cookie attributes as a TypedDict
`utils/oauth.py:58` infers `dict[str, object]` for `_NONCE_COOKIE_ATTRS`, which produces 8 mypy `arg-type` errors where it is unpacked into `set_cookie` and `delete_cookie`.

CI carries them because its mypy step is `continue-on-error: true`, but a local pre-commit hook running mypy blocks every commit on the file. A `TypedDict` annotation fixes it without changing the emitted cookie.

jisung-02:
## Background

`_NONCE_COOKIE_ATTRS` in `utils/oauth.py` is a bare dict literal, so mypy infers its values as `dict[str, object]`. Unpacking it in `_set_nonce_cookie` and `_clear_nonce_cookie` produces one `arg-type` error per key (`path`, `secure`, `httponly`, `samesite`) at each of the two call sites, eight in total.

The mypy step in `.github/workflows/test.yml:55` is `continue-on-error: true`, so CI stays green. A local pre-commit hook running mypy is not, and it blocks every commit that touches the file.

## Changes

Added a `_NonceCookieAttrs` TypedDict and annotated `_NONCE_COOKIE_ATTRS` with it. `samesite` is narrowed to the `Literal['lax', 'strict', 'none']` starlette expects.

`path` and `secure` are pinned to `Literal['/']` and `Literal[True]`, following review. The `__Host-` prefix makes the browser require both, and that invariant previously lived only in the comment below the constant, so `'secure': False` type-checked clean. `httponly` stays `bool` because `__Host-` does not require HttpOnly — that one is our own policy, not a browser rule.

No call site changes: `Literal[True]` is a subtype of `bool` and `Literal['/']` of `str`, so both still satisfy starlette's `set_cookie` and `delete_cookie` signatures. The dict values are untouched, so the emitted Set-Cookie header is unchanged.

The class sits above the comment rather than below it: the comment explains why set and delete share these values, so it belongs with the constant.

## Testing

- `mypy --ignore-missing-imports --no-strict-optional .`: `Success: no issues found in 82 source files`, down from `Found 8 errors in 1 file`
- Negative check on the pinned values: setting `'path': '/oauth'` and `'secure': False` now fails with two `typeddict-item` errors; reverting them is clean again
- `ruff check .`: `All checks passed!`
- `pytest tests/`: `1125 passed, 7 warnings in 7.21s`

Closes #167

### DIFF
--- a/utils/oauth.py
+++ b/utils/oauth.py
@@ -19,6 +19,7 @@ import secrets
 import time
 from functools import wraps
 from http import HTTPStatus
+from typing import Literal, TypedDict
 from urllib.parse import parse_qs, urlencode, urlparse, urlunparse
 
 import httpx
@@ -51,11 +52,19 @@ _STATE_TTL_SECONDS = 600
 # Domain, so the cookie cannot be planted by a sibling host.
 _NONCE_COOKIE_NAME = '__Host-alpacon_oauth_nonce'
 
+
+class _NonceCookieAttrs(TypedDict):
+    path: str
+    secure: bool
+    httponly: bool
+    samesite: Literal['lax', 'strict', 'none']
+
+
 # Shared by set and delete: the delete has to repeat Path, or it addresses a
 # different cookie, and Secure, or the browser rejects the whole Set-Cookie
 # under a __Host- name and the cookie survives. SameSite stays Lax because
 # Strict drops the cookie on the top-level return from Auth0.
-_NONCE_COOKIE_ATTRS = {
+_NONCE_COOKIE_ATTRS: _NonceCookieAttrs = {
     'path': '/',
     'secure': True,
     'httponly': True,
--- a/utils/oauth.py
+++ b/utils/oauth.py
@@ -54,8 +54,8 @@ _NONCE_COOKIE_NAME = '__Host-alpacon_oauth_nonce'
 
 
 class _NonceCookieAttrs(TypedDict):
-    path: str
-    secure: bool
+    path: Literal['/']
+    secure: Literal[True]
     httponly: bool
     samesite: Literal['lax', 'strict', 'none']
 


utils/oauth.py
```diff
@@ -51,11 +52,19 @@
 # Domain, so the cookie cannot be planted by a sibling host.
 _NONCE_COOKIE_NAME = '__Host-alpacon_oauth_nonce'
 
+
+class _NonceCookieAttrs(TypedDict):
+    path: str
+    secure: bool
```
junho226: Optional, and arguably out of scope for this PR — flagging it because the comment four lines below already states the invariant in prose.

`secure: bool` and `path: str` accept any value of that type: I confirmed that flipping the constant to `'secure': False` still type-checks clean under the CI mypy flags. Since the `__Host-` prefix makes `Secure` and `Path=/` mandatory rather than merely conventional, `Literal` would let the type carry that:

```python
class _NonceCookieAttrs(TypedDict):
    path: Literal['/']
    secure: Literal[True]
    httponly: bool
    samesite: Literal['lax', 'strict', 'none']
```

`Literal` is already imported, so this costs nothing extra. It does go beyond the eight `arg-type` errors this PR set out to fix, so leaving it as-is and filing a follow-up is an equally reasonable call.

junho226 (dismissed):
## Summary

A 10-line typing change on `utils/oauth.py`: `_NONCE_COOKIE_ATTRS` gains a `_NonceCookieAttrs` TypedDict annotation, with `samesite` narrowed to the `Literal['lax', 'strict', 'none']` starlette declares. No dict value moves.

I treated this as a "pure typing change that could quietly alter cookie semantics" and checked it that way: attribute-by-attribute value comparison against `main`, the starlette parameter names the dict is unpacked into, and set/delete symmetry. It holds — the emitted `Set-Cookie` header is unchanged.

## Critical issues

None.

## Warnings

None.

## Suggestions

All three are optional and none belongs in this PR as written.

1. **`secure` and `path` could be pinned by the type, not only by convention.** `secure: bool` and `path: str` accept anything of that type. I confirmed `'secure': False` still type-checks clean. `Literal[True]` and `Literal['/']` would put the two `__Host-` invariants the comment above the constant already explains into the type itself. Inline note on the class body. Deliberately out of scope for a fix aimed at eight `arg-type` errors — raise it only if you want the invariant enforced rather than documented.

2. **The clear path has no attribute-level test.** `test_authorize_sets_the_nonce_cookie` (`tests/test_oauth.py:882-887`) pins `secure`, `httponly`, `samesite=lax`, `path=/` and `max-age=600` on the set header. The two clear tests (`test_final_redirect_clears_the_cookie:1996`, `test_json_fallback_clears_the_cookie:2009`) assert only the cookie name and `max-age=0` / `expires=`. Since a delete that drops `Path` or `Secure` silently fails to clear the cookie — exactly what the shared constant exists to prevent — asserting `path=/` and `secure` on the delete header too would pin the symmetry. Follow-up issue, not a change to this PR.

3. **This makes the whole repo mypy-clean, which opens a follow-up.** `mypy --ignore-missing-imports --no-strict-optional .` now reports `Success: no issues found in 82 source files`. `.github/workflows/test.yml` still carries `continue-on-error: true` on the mypy step. With zero errors left repo-wide, dropping that flag would turn the type check into a real gate and stop the next regression from landing the same way. Separate PR.

## Highlights

- **Behavior-preserving, verified rather than assumed.** I parsed both revisions with `ast.literal_eval` and compared the constants: base and head both yield `{'path': '/', 'secure': True, 'httponly': True, 'samesite': 'lax'}`, `IDENTICAL: True`, with `_NONCE_COOKIE_NAME` unchanged at `__Host-alpacon_oauth_nonce`. No key added, dropped, or renamed.
- **`__Host-` invariants intact.** `Secure` is `True`, `Path` is `/`, and no `domain` key exists — all three still hold.
- **Set/delete symmetry survives, on both call sites.** `_set_nonce_cookie` (`utils/oauth.py:459-460`) and `_clear_nonce_cookie` (`utils/oauth.py:475`) both unpack the same `**_NONCE_COOKIE_ATTRS`. Against starlette 1.6.0, `set_cookie` and `delete_cookie` each declare `path`, `domain`, `secure`, `httponly`, `samesite` — every TypedDict key maps to a real parameter of both, so nothing lands in the wrong slot. `max_age` correctly stays out of the shared dict: `set_cookie` passes it explicitly, and `delete_cookie` has no such parameter (it hardcodes `max_age=0, expires=0` internally).
- **The typing is load-bearing, not decorative.** The dict is never widened to `dict[str, Any]`, and mypy checks the literal against the TypedDict. Confirmed with three deliberate mutations, each reverted:
  - `'samesite': 'bogus'` → `error: Incompatible types (expression has type "Literal['bogus']", TypedDict item "samesite" has type "Literal['lax', 'strict', 'none']") [typeddict-item]`
  - adding `'domain': ...` → `error: Extra key "domain" for TypedDict "_NonceCookieAttrs" [typeddict-unknown-key]`
  - removing `'path'` → `error: Missing key "path" for TypedDict "_NonceCookieAttrs" [typeddict-item]`

  The middle one is a small bonus: because the TypedDict is total, planting a `Domain` on a `__Host-` cookie is now a type error rather than a silent security regression.
- **`total=True` (the default) is the right choice here.** All four keys are required, which is precisely what mypy needs to accept a `**` unpack into a call; `total=False` or `NotRequired` would have reintroduced the errors this PR removes.
- The PR body's reasoning for placing the class above the comment rather than below it is correct — the comment explains the values, so it belongs against the constant.

## Verification

Run in a clean worktree at `045c9a5`, with `uv venv` + `uv pip install -e .[dev]` (starlette 1.6.0, Python 3.12.9). Commands mirror `.github/workflows/test.yml`.

| Command | Result |
|---|---|
| `uv run --no-sync ruff check .` | pass — `All checks passed!` |
| `uv run --no-sync ruff format . --check` | pass — `96 files already formatted` |
| `uv run --no-sync mypy --ignore-missing-imports --no-strict-optional .` (head) | pass — `Success: no issues found in 82 source files` |
| same command with `utils/oauth.py` reverted to `origin/main` | `Found 8 errors in 1 file (checked 82 source files)` — 4 `arg-type` at the `set_cookie` call, 4 at `delete_cookie`, reproducing the issue exactly |
| `PYTHONPATH=. uv run --no-sync pytest tests/ -q` | pass — `1125 passed, 8 warnings in 7.89s` |
| AST value diff of `_NONCE_COOKIE_ATTRS`, base vs head | `IDENTICAL: True` |
| 3 deliberate-mutation mypy runs (bad `samesite`, extra `domain`, missing `path`) | each errored as quoted above; working tree restored, `git status --porcelain` empty |

The 8 warnings in the pytest run are pre-existing `RuntimeWarning: coroutine ... was never awaited` in `tests/test_system_tools.py`, untouched by this PR. The PR body reports 7; the difference is environmental, not a change in behavior.

Not run: nothing. Every check the CI workflow performs was executed locally, on all three of ruff, mypy and pytest. CI runs the matrix on Python 3.12/3.13/3.14 and I only exercised 3.12; the change uses no version-dependent syntax (`typing.TypedDict` and `typing.Literal` are stable across all three), so I do not consider that a gap.

## Verdict

**Approve.** The change does exactly what it claims: eight mypy errors removed, zero cookie semantics touched, both consumers of the shared constant still consuming it. Values verified identical against `main`, the `__Host-` invariants verified intact, and the new type verified to actually reject bad values rather than pass them through.

## Checklist

- [x] Behavior-preserving — attribute dict compared key by key against `main`, identical
- [x] `httponly` / `secure` / `samesite` / `path` unchanged; no `domain`, no `max_age` in the shared dict
- [x] `__Host-` requirements hold: `Secure` present, `Path=/`, no `Domain`
- [x] TypedDict keys match the starlette parameter names of both `set_cookie` and `delete_cookie`
- [x] `total=True` correct for `**` unpacking; no `NotRequired`, no widening to `dict[str, Any]`
- [x] Set and delete both consume the shared value — symmetry preserved
- [x] Browser-binding property (nonce hash in state, HMAC compare, `__Host-` cookie) untouched
- [x] Tests pin the attribute values on the set path, not just the types
- [x] ruff, mypy, pytest run locally and observed passing
- [x] Prior review read — Copilot posted a descriptive overview with no findings; nothing to re-litigate

utils/oauth.py
jisung-02: Applied in e3facf1 — `path: Literal['/']` and `secure: Literal[True]`.

I verified the narrowing actually bites rather than assuming it: flipping the constant to `'path': '/oauth'` and `'secure': False` now fails with two `typeddict-item` errors under the CI mypy flags, and reverting them is clean again. Full run is `Success: no issues found in 82 source files`, `pytest` `1125 passed`.

`httponly` stays `bool`. `__Host-` requires Secure and `Path=/` but says nothing about HttpOnly, so pinning that one would encode our own policy as a browser rule.

No call site changes were needed, since `Literal[True]` is a subtype of `bool` and `Literal['/']` of `str`.

junho226 (approved):
## Round 2

The delta since my approval at `045c9a5` is a single commit, `e3facf1`, changing two lines in `utils/oauth.py`: `path: str` to `Literal['/']` and `secure: bool` to `Literal[True]`. Nothing else moved — `git diff 045c9a5..HEAD` is those four lines, and `tests/` is untouched by both the delta and the PR as a whole.

| Round-1 point | Status |
|---|---|
| 1. Pin `secure` / `path` with `Literal` | **Fixed** in `e3facf1` |
| 2. No attribute-level test on the clear path | **Still open**, correctly deferred — I had marked it "follow-up issue, not a change to this PR" |
| 3. Drop `continue-on-error` on the mypy step | **Still open**, correctly deferred — I had marked it "separate PR" |

**1 — fixed, and I verified the narrowing rather than taking the reply at its word.** Reproduced from scratch at head: `'secure': False` now gives `error: Incompatible types (expression has type "Literal[False]", TypedDict item "secure" has type "Literal[True]")  [typeddict-item]`, and `'path': '/oauth'` gives the matching `Literal['/oauth']` against `Literal['/']`. The counterfactual is the part that actually settles it — the same `'secure': False` mutation applied at `045c9a5` still reports `Success: no issues found`, so the delta closes a gap that was genuinely open rather than restating one that was already covered.

The reasoning for leaving `httponly: bool` alone is right, and it holds in practice: mutating `'httponly': False` at head still type-checks clean. `__Host-` constrains `Secure`, `Path` and `Domain` and says nothing about HttpOnly, so pinning it would have encoded a house policy as a browser rule. I would not want it pinned either.

**2 — still open, and that is the right outcome.** `tests/test_oauth.py` is unchanged. The set-path test still pins the attributes (`test_authorize_sets_the_nonce_cookie:875`, asserting `secure`, `httponly`, `samesite=lax`, `path=/`, `max-age=600`), while the two clear-path tests (`test_final_redirect_clears_the_cookie:1996`, `test_json_fallback_clears_the_cookie:2009`) still assert only the cookie-name prefix and `max-age=0` / `expires=`. That was explicitly a follow-up, not a condition of this PR, so it is not a reason to hold the merge. Worth noting only that no follow-up issue exists yet — the open OAuth follow-up, #160, covers per-flow cookie naming, not this.

**3 — still open, and that is also right.** `.github/workflows/test.yml:55` still carries `continue-on-error: true` on the mypy step, and the premise for dropping it still holds at head: `Success: no issues found in 82 source files`. Separate PR, as stated.

## New findings

None.

One thing I checked specifically before dismissing it: `Literal[True]` on `secure` means loosening the cookie for a plain-HTTP local flow now takes a type edit, not just a value edit. That is not a regression. The value was already hardcoded `True` on `main`; there is no dev or insecure toggle anywhere in the repo (grep for `dev_mode`, `insecure`, `allow_http`, `NO_TLS` across the tree returns nothing); and the constant has exactly two consumers, both unpacking it wholesale with no runtime variation path. A plain-HTTP flow would fail on the `__Host-` cookie name well before it reached the `Secure` attribute, so the type now pins something the cookie name already required.

## Verification

Fresh worktree at `e3facf1`, `uv venv` plus `uv pip install -e '.[dev]'` (starlette 1.6.0, Python 3.12.9). Commands mirror `.github/workflows/test.yml`.

| Command | Result |
|---|---|
| `uv run --no-sync ruff check .` | pass — `All checks passed!` |
| `uv run --no-sync ruff format . --check` | pass — `96 files already formatted` |
| `uv run --no-sync mypy --ignore-missing-imports --no-strict-optional .` | pass — `Success: no issues found in 82 source files` |
| `PYTHONPATH=. uv run --no-sync pytest tests/ -q` | pass — `1125 passed, 8 warnings in 7.98s` |

Re-measured at head rather than carried forward, because the delta touched the constant's type:

| Check | Result |
|---|---|
| AST `literal_eval` of `_NONCE_COOKIE_ATTRS`, `origin/main` vs head | `{'path': '/', 'secure': True, 'httponly': True, 'samesite': 'lax'}` on both sides — `IDENTICAL: True` |
| same, `045c9a5` vs head | `IDENTICAL: True` — the delta moved no value |
| `_NONCE_COOKIE_NAME`, `origin/main` vs head | identical, `__Host-alpacon_oauth_nonce` |
| `__Host-` invariants at head | `Secure` is `True`, `Path` is `/`, no `domain` key, name carries the prefix — all three hold |
| TypedDict keys vs dict keys | equal, all four present; no `total=False`, no `NotRequired`, no base beyond `TypedDict` |
| TypedDict keys vs starlette 1.6.0 parameters | every key is a real parameter of both `set_cookie` and `delete_cookie`; `max_age` is accepted by `set_cookie` and absent from `delete_cookie`, so keeping it out of the shared dict is still correct |
| Set/delete symmetry | `_set_nonce_cookie:459-460` and `_clear_nonce_cookie:475` remain the only two consumers, both unpacking `**_NONCE_COOKIE_ATTRS` |
| Runtime types of the values | `str` / `bool` — `Literal` narrows statically only, so the emitted `Set-Cookie` header is unchanged |

Mutation runs, each reverted afterwards:

| Mutation | Result |
|---|---|
| head, `'secure': False` | `typeddict-item` error — `Literal[False]` against `Literal[True]` |
| head, `'path': '/oauth'` | `typeddict-item` error — `Literal['/oauth']` against `Literal['/']` |
| head, `'samesite': 'invalid'` | `typeddict-item` error — against `Literal['lax', 'strict', 'none']` |
| head, `'httponly': False` | clean, as intended — the narrowing is deliberately scoped |
| `045c9a5`, `'secure': False` | clean — confirms the gap the delta closes was real |

Working tree restored after every mutation: `git status --short` empty, `git diff HEAD` empty, and `utils/oauth.py` byte-identical to the committed blob.

Carried forward from round 1 and not re-run: the `origin/main` reproduction of the original eight `arg-type` errors, and the `domain`-added / `path`-removed mutations. Both concern code the delta does not touch.

Not run: the Python 3.13 and 3.14 legs of the CI matrix — I exercised 3.12 only. `typing.TypedDict` and `typing.Literal` behave identically across the three and the change adds no version-dependent syntax, so I do not count that as a gap.

## Verdict

**Approve**, unchanged from round 1. The one point worth acting on inside this PR was acted on and independently verified; the two follow-ups remain follow-ups. The delta moves no cookie value, so `main` and head still emit byte-identical `Set-Cookie` attributes.

PR #181, opened 2026-08-14
feat(oauth): validate ALPACON_MCP_STATE_SECRET at startup

### LINKED ISSUE #156: [FEATURE] Validate ALPACON_MCP_STATE_SECRET at startup instead of at first sign
## Background

PR #152 added a hex-format and 32-byte minimum check to `ALPACON_MCP_STATE_SECRET`. The check lives inside `_get_state_secret()`, which is only called when a state is signed or verified, so a malformed key surfaces at the **first OAuth request** after the server is already up.

`d8c51b8` made that request return the diagnostic instead of a bare 500, but it did not change *when* the problem is found. The deployment still looks successful, and the first user who tries to authenticate is the one who hits the failure.

Raised by @junho226 in the re-review of #152 (`utils/oauth.py:871`). It was kept out of that PR because commit `f34759a` recorded "validation runs on the existing lazy path" as a `Constraint`; moving it to startup reverses that decision, which is a call worth making separately.

## Proposal

`register_oauth_routes` (around `utils/oauth.py:871`) already reads `os.getenv(_STATE_SECRET_ENV)` for a log line. One `_get_state_secret()` call there turns the lazy per-request 500 into a boot-time failure that carries the message.

```python
# rough shape; confirm against the current code when implementing
if os.getenv(_STATE_SECRET_ENV):
    _get_state_secret()  # raises ValueError on a malformed value
```

## Open questions

1. **Whether to validate the derived path too.** With no explicit key, `_get_state_secret()` falls through to `_get_oauth_config()` and reads `AUTH0_CLIENT_SECRET`. Calling it unconditionally would break deployments that currently start without OAuth configured, so gating on the `os.getenv` result is the safe form.
2. **Fail to boot, or log a warning.** A server running with a malformed key returns 500 on every OAuth request anyway, so failing fast looks right — but this is an operations policy call.
3. **Import-time side effects.** `utils/oauth.py` deliberately does not read OAuth config at import time. `register_oauth_routes` is call-time, so it preserves that property; moving the check any higher would not.

## Testing

- `register_oauth_routes` raises `ValueError` with the message when `ALPACON_MCP_STATE_SECRET` is malformed.
- Registration still succeeds when `ALPACON_MCP_STATE_SECRET` is unset and OAuth config is incomplete (regression guard for open question 1).

## References

- PR #152 (`f34759a`, `d8c51b8`)
- Review comment: https://github.com/alpacax/alpacon-mcp/pull/152#discussion_r3709905371

jisung-02:
## Background

PR #152 added the hex-format and 32-byte minimum checks for `ALPACON_MCP_STATE_SECRET`. They live inside `_get_state_secret()` (`utils/oauth.py:372`), which is only called from `_sign_state` and `_verify_state`, so nothing reads the variable until an OAuth request arrives.

A deployment that sets a malformed value therefore boots, logs `OAuth proxy routes registered`, and reports success. The failure appears on the first authentication attempt, found by a user rather than by whoever deployed it. Commit `d8c51b8` made that request return the diagnostic instead of a bare 500, but the timing was unchanged.

Raised by @junho226 in the re-review of #152 (https://github.com/alpacax/alpacon-mcp/pull/152#discussion_r3709905371). It was left out of that PR because `f34759a` had recorded "validation runs on the existing lazy path" as a `Constraint`; this PR is where that call gets made again.

## Changes

`register_oauth_routes` calls `_get_state_secret()` once, at the top of the function:

```python
if os.getenv(_STATE_SECRET_ENV):
    _get_state_secret()
```

A malformed value now raises `ValueError` with the same message during route registration, which `server.py:358` runs at startup. The call is placed before the first `@mcp_server.custom_route` rather than beside the existing log line at the end, so a failure cannot leave the server holding a partially registered route table.

The `os.getenv` gate is the answer to open question 1 in the issue. Without an explicit key, `_get_state_secret()` falls through to `_get_oauth_config()` and reads `AUTH0_CLIENT_SECRET`; calling it unconditionally would fail the boot of any deployment that currently runs with OAuth unconfigured.

Open question 2 — fail to boot, or warn — is resolved as failing to boot. A server holding a malformed key answers every OAuth request with a 500, so a warning would only postpone the same failure to the same place it happens today.

Open question 3 stands unchanged: `register_oauth_routes` is call-time, so `utils/oauth.py` still reads no OAuth config at import time.

`docs/configuration.md:220` said a bad value "is rejected when the OAuth endpoints first sign or verify a state", which this change makes wrong. It now says the server fails at startup.

## Testing

Both tests from the issue's Testing section, added to `TestStateSecret` in `tests/test_oauth.py`:

- `test_registration_rejects_malformed_explicit_secret`: `register_oauth_routes` raises `ValueError` matching `must be hex`. Run against the unmodified `utils/oauth.py` it fails with `Failed: DID NOT RAISE <class 'ValueError'>`, so it covers the change rather than passing either way.
- `test_registration_survives_incomplete_oauth_config`: registration still succeeds with `ALPACON_MCP_STATE_SECRET` and `AUTH0_CLIENT_SECRET` both empty — the regression guard for open question 1.

Commands run on the final tree:

- `uv run python -m pytest tests/`: `1127 passed, 7 warnings in 8.84s`
- `uv run ruff check .`: `All checks passed!`
- `uv run ruff format --check utils/oauth.py tests/test_oauth.py`: `2 files already formatted`

End to end, running `python main_http.py` locally with placeholder Auth0 values:

| `ALPACON_MCP_STATE_SECRET` | Result |
|---|---|
| `correct-horse-battery` | exit 1, `ValueError: ALPACON_MCP_STATE_SECRET must be hex; generate one with openssl rand -hex 32` |
| 16 bytes of hex | exit 1, `ValueError: ALPACON_MCP_STATE_SECRET must decode to at least 32 bytes` |
| 32 bytes of hex | boots, `state secret: explicit env` |
| unset | boots, `state secret: derived from client secret` |

The traceback on the failing cases runs `server.py:358 register_oauth_routes` to `utils/oauth.py:492 _get_state_secret()`, so the check fires during registration. With the parent commit's `utils/oauth.py` restored, the same malformed value reaches `Uvicorn running on http://127.0.0.1:8933` and serves.

Not verified: no server was booted against a real Auth0 tenant, and the 400 that `GET /authorize` returned on the parent-commit run was not traced to a specific handler, so it is not evidence of the per-request failure path.

Closes #156


utils/oauth.py
```diff
@@ -480,7 +480,16 @@ def register_oauth_routes(mcp_server):
...
         mcp_server: FastMCP server instance
+
+    Raises:
+        ValueError: ALPACON_MCP_STATE_SECRET is set to a malformed value.
     """
+    # A malformed key otherwise surfaces on the first user's OAuth request,
+    # long after the deployment reported success. Only the explicit key is
+    # checked: the derived path needs OAuth config, and deployments that run
+    # without it have to keep starting.
+    if os.getenv(_STATE_SECRET_ENV):
```
junho226, line 491: Nit, non-blocking. This truthiness test is now the third copy of the same predicate: `_get_state_secret()` uses it at `utils/oauth.py:377-378` to pick the explicit path, the registration log at `utils/oauth.py:1252` uses it to report the source, and this gate uses it to decide whether to validate. They agree today, and I confirmed the gate cannot decline to check a value the signer would accept.

The coupling is only worth a mention because the three must stay in lockstep: if the explicit path ever gained normalization — a `.strip()`, say — a value like `'  \t  '` would become falsy for `_get_state_secret()` while staying truthy here, and the gate would start validating a value the signer treats as unset. A shared `_has_explicit_state_secret()` predicate would remove that possibility. Fine to leave as is if you prefer the explicit `os.getenv` at each site.

junho226 (approved):
## Review: #181 — validate `ALPACON_MCP_STATE_SECRET` at startup

**Stack**: python (`CLAUDE.md` frontmatter) | **Changes**: +24 -1 in 3 files | **Branch**: `156-validate-state-secret-at-startup` → `main`

### Summary

Moves the hex/32-byte validation of `ALPACON_MCP_STATE_SECRET` from the lazy `_sign_state`/`_verify_state` path to route registration, so a malformed key fails the boot instead of the first user's OAuth request. Two lines of production code, gated on `os.getenv(_STATE_SECRET_ENV)` so a deployment running with OAuth unconfigured keeps starting.

The change is correct and fail-closed. I verified every load-bearing claim in the description against the code rather than reading it off the body.

### Verification

**Placement and fail-closed behavior.** The check sits at `utils/oauth.py:491`, ahead of the first `@mcp_server.custom_route` at `utils/oauth.py:496`. I exercised registration directly against a `MockMCPServer` across the input space, counting routes registered on the failure path:

| `ALPACON_MCP_STATE_SECRET` | Result | Routes registered |
|---|---|---|
| unset | boots, `derived from client secret` | 8 |
| `''` (empty) | boots, `derived from client secret` | 8 |
| `'   '` (whitespace) | `ValueError: … must decode to at least 32 bytes` | 0 |
| `correct-horse-battery` | `ValueError: … must be hex` | 0 |
| `abc` (odd length) | `ValueError: … must be hex` | 0 |
| `0x`-prefixed 32 bytes | `ValueError: … must be hex` | 0 |
| 31 bytes of hex | `ValueError: … must decode to at least 32 bytes` | 0 |
| 32 bytes of hex | boots, `explicit env` | 8 |

`routes=0` on every rejection confirms the stated reason for putting the call at the top of the function: the failure cannot leave a partially registered route table. Whitespace-only is worth calling out because it is the one input that reaches `bytes.fromhex` and does *not* raise there — `fromhex` skips ASCII whitespace and returns `b''` — and the length check catches it. Fail-closed either way.

**Startup, not first request.** `register_oauth_routes(mcp)` runs at `server.py:358`; `mcp.run(transport=transport)` is at `server.py:384`, so the raise precedes listening. `main_http.py:50` forces `ALPACON_MCP_AUTH_ENABLED=true`, so the `remote_mode` branch that reaches the check is unconditional for the HTTP entry point, and `Dockerfile:26` (`CMD ["python", "main_http.py"]`) is that entry point in the deployed image. `main_http.py:62-64` logs and re-raises, so the process exits non-zero. No bypass in a test or dev path: stdio and SSE skip registration only when auth is disabled, and those transports never sign a state.

**No secret in logs or errors.** Both `ValueError` messages name the variable and the constraint and never interpolate the value; I asserted this per case (`value_leaked_in_msg=False` for all of them). `main_http.py:63` re-logs `{e}` with `exc_info=True`, and the standard traceback renders source lines, not frame locals, so the raw value stays out of the log.

**The gate matches the consumer.** `if os.getenv(_STATE_SECRET_ENV)` is the same truthiness test `_get_state_secret()` applies at `utils/oauth.py:377-378`, so there is no window where the gate declines to check a value the signer would then accept. Empty string takes the derived path in both.

**Tests.** `test_registration_rejects_malformed_explicit_secret` fails on the parent tree — I restored `utils/oauth.py` from the merge base (`02894ea`) and got `Failed: DID NOT RAISE ValueError`, then restored the branch version. The test covers the change rather than passing either way, as claimed.

**Docs.** `docs/configuration.md:220` was the only prose asserting the old lazy timing; the other mention (`:278`) is the `openssl rand -hex 32` snippet and stays accurate. No stale claim left behind.

### Commands run

On the branch tree, in a fresh venv, using the commands from `.github/workflows/test.yml`:

- `uv run --no-sync ruff check .` — `All checks passed!`
- `uv run --no-sync ruff format --check utils/oauth.py tests/test_oauth.py` — `2 files already formatted`
- `uv run --no-sync mypy --ignore-missing-imports --no-strict-optional .` — `Success: no issues found in 82 source files`
- `PYTHONPATH=. uv run --no-sync pytest tests/ -q` — `1127 passed, 8 warnings in 7.22s`

### Not verified

- No server was booted against a real Auth0 tenant; the end-to-end table in the description is not independently reproduced here. The registration-level equivalent is the matrix above, which exercises the same call.
- CI runs the mypy step with `continue-on-error: true`, so a clean mypy is informational rather than a gate.

### Notes

One non-blocking nit inline about the truthiness test now living in three places. Nothing to fix before merge.

**Verdict**: APPROVE

- [x] Follows the repo's conventions (comments explain the why, `Raises:` documented, `os` already imported)
- [x] Tests included, and shown to fail without the change
- [x] No secret exposure in errors or logs; failure is fail-closed at startup
- [x] Docs updated, no stale statement of the old behavior

utils/oauth.py
jisung-02: Leaving it as is, with one note on the shape.

A boolean `_has_explicit_state_secret()` would only cover this gate and the registration log at `utils/oauth.py:1252`. `_get_state_secret()` needs the value, not the predicate — `explicit = os.getenv(_STATE_SECRET_ENV, '')` at `utils/oauth.py:377` — so it would keep its own read, and the pair that has to stay in lockstep is exactly the gate and the signer. Closing it properly means a value-returning accessor that all three go through, not a boolean one.

That is worth doing the day the explicit path gains normalization. Until then the predicate is a bare `os.getenv` in all three places and cannot drift on its own, so I would rather not add the indirection now.

PR #183, opened 2026-08-14
refactor: replace HTTP status code magic numbers with http.HTTPStatus

### LINKED ISSUE #143: refactor: replace HTTP status code magic numbers with http.HTTPStatus
## Problem

HTTP status codes are scattered across the codebase as numeric literals. Only `tools/webftp_tools.py` uses `http.HTTPStatus`, so two conventions coexist in the same repository.

The single place that uses `http.HTTPStatus` today:

| Location | Code |
|------|------|
| `tools/webftp_tools.py:8` | `from http import HTTPStatus` |
| `tools/webftp_tools.py:47` | `_S3_SUCCESS_CODES = (HTTPStatus.OK, HTTPStatus.CREATED)` |
| `tools/webftp_tools.py:78` | `response.status_code != HTTPStatus.OK` |
| `tools/webftp_tools.py:143`, `840` | `resp.status_code (not) in _S3_SUCCESS_CODES` |

Everything else is a magic number:

| File | Literals | Examples |
|------|------|------|
| `utils/oauth.py` | 25 | `status_code=400`, `status_code=302`, `status_code=502`, `status_code=201`, `response.status_code == 200` (line 429) |
| `utils/http_client.py` | 4 | `e.response.status_code != 401` (390), `>= 500` (393), `== 401` (405), `'status_code': 401` (233) |
| `utils/auth_error_middleware.py` | 2 | `status_code=500` (176), `'status': 401` (302) |
| `tools/workspace_tools.py` | 2 | `result.get('status_code') == 404` (88), `status_code=404` (95) |

## Secondary problem: inconsistent key types in status code tables

Two modules hold what is essentially the same status-code-to-message mapping, but with different key types.

- `utils/error_handler.py:199-227`: string keys `'400'` through `'503'`, with comparisons like `error_code == '404'` (lines 251, 253)
- `utils/recovery_hints.py:8-88`: integer tuple keys such as `(401, 'general')` and `(403, 'command')`

As a result, `recovery_hints._parse_status_code()` has to normalize the value with `int()` separately.

## Proposal

Standardize on `http.HTTPStatus`, not `httpx.codes`:

- It is stdlib, so no dependency spreads. httpx is currently imported only by `utils/http_client.py` and `tools/webftp_tools.py`; there is no reason to pull httpx into `utils/oauth.py` (which deals with Starlette responses) or into `utils/error_handler.py` and `utils/recovery_hints.py`, which are unrelated to httpx, just to get status code constants.
- `tools/webftp_tools.py` already uses it, so adopting it means aligning the rest without touching existing code.
- The only codes needed here are 200/201/302/400/401/403/404/429/500/502/503, all covered by `HTTPStatus`.

`requires-python = ">=3.12"`, so the `IntEnum.__str__` change from 3.11 (where `f'{HTTPStatus.OK}'` used to render as `'HTTPStatus.OK'` on 3.10 and below) does not apply. On 3.12 it formats as `'200'`.

## Prerequisite decision

The `error_messages` table in `utils/error_handler.py` needs a design decision before its keys can become ints. The same dict also holds non-numeric keys (`'timeout'`, `'network'`, `'validation'`), so we either allow mixed `int | str` keys or split the table into numeric and non-numeric parts. This should be settled before the refactor starts.

## Scope

- Replace status code literals in the four files above with `HTTPStatus` members
- Unify the key types of the `error_handler` and `recovery_hints` tables (after the prerequisite decision)
- Out of scope: three-digit numbers that are not status codes, including `utils/http_client.py:34` `default_cache_ttl = 300`, `:52` `max_connections=100`, `utils/security_settings.py:21` `_CACHE_TTL = 300`, `tools/command_tools.py:224` `timeout = 300`, `tools/metrics_tools.py:889` `hours > 168`, `tools/security_tools.py:22` `_BULK_MAX_SERVERS = 100`, `tools/system_info_tools.py:231` `limit = 100`
- No behavior change. Existing tests must pass as is; literals in test code may stay as they are

jisung-02:
## Background

Closes #143

HTTP status codes were written as bare numeric literals (`401`, `404`, `500`) across several production modules, while `tools/webftp_tools.py` already used `http.HTTPStatus`. This PR removes that split and standardizes on `HTTPStatus` in both the production code and the test suite.

## Changes

Production: converted 29 sites in `utils/oauth.py`, 11 in `utils/recovery_hints.py`, 4 in `utils/http_client.py`, 3 in `utils/auth_error_middleware.py`, and 2 in `tools/workspace_tools.py`. In `recovery_hints.py` the `_HINT_REGISTRY` keys now read as `(HTTPStatus.NOT_FOUND, 'server')`, so the 11 grouping comments (`# 404 errors` and friends) that only restated the key were deleted.

Tests: converted 237 literals across 23 files, so an expected status names the same constant the production code produces. Beyond the `status_code = 404` and `'status_code': 404` shapes, this covers the positional forms — `httpx.Response(404, ...)`, `get_recovery_hints(403, ...)`, `_detect_error_domain(404, ...)` — and the ASGI `'status': 401` field in `tests/test_auth_error_middleware.py`.

Status codes inside string literals stay as strings. The mocked exception message `'HTTP 500 Internal Server Error'` and the assertion `'404' in result['message']` both exercise rendered text, so the digits are the subject there rather than a magic number. The same applies to the `'404'` input in `tests/test_recovery_hints.py`, which tests string-to-int parsing.

`format_user_friendly_error` in `utils/error_handler.py` is removed. The issue named it as a prerequisite decision ("string keys or int keys?"), but nothing calls it. Verified five ways: static grep, dynamic-dispatch grep (`getattr`/`globals()`), runtime test coverage, git history of its former callers, and the packaging export surface. The module-level `logger` went with it — that function's `logger.debug` was its only consumer.

## No behavior change

`HTTPStatus` is an `IntEnum`, so `==`, `>=`, and `hash()` behave exactly as `int`. Dict keys, comparisons, and JSON serialization are unaffected. On Python 3.12 `str()` and f-string interpolation render `'401'`, not the pre-3.11 `'HTTPStatus.UNAUTHORIZED'` form.

The one site worth scrutiny is `auth_error_middleware.py` putting an `IntEnum` into the ASGI `http.response.start` `'status'` field, so it was measured rather than assumed. On the h11 path the wire bytes are byte-identical (`b'HTTP/1.1 401 \r\n...'`); on the httptools path the value is only used for a `STATUS_LINE[status_code]` dict lookup and a `status_code not in (204, 304)` check, both of which resolve correctly for an `IntEnum`. Measured against uvicorn 0.34.3 and h11 0.16.0.

Stdlib `http.HTTPStatus` was chosen over `httpx.codes` so the httpx dependency does not spread into the Starlette/OAuth layer or into non-HTTP modules like `utils/recovery_hints.py`.

## Testing

`pytest` reports 1127 passed, the same count as before the sweep. `ruff check .` passes and `ruff format --check .` reports no reformatting needed.

Completeness was established with the Python tokenizer rather than grep, since a regex pass over `status_code`/`status` misses the positional call forms. Walking every `NUMBER` token in every `.py` file leaves four numbers in the 2xx-5xx range, none of them a status code: `peak_output_pps`, an `hours` argument, a result `count`, and a `text[:200]` slice.

## Notes for the reviewer

The line numbers and counts in the issue body are from 2026-07-28 and no longer match: `oauth.py` had 27 `status_code=` sites plus 2 comparisons rather than 25, `auth_error_middleware.py` had 3 rather than 2, and `204` was missing from the issue entirely.

`utils/oauth.py:1126` has a comment reading "Log non-2xx responses" above a condition of `>= HTTPStatus.BAD_REQUEST`, which skips 3xx. That mismatch predates this PR and is left as-is.

### DIFF
--- a/tests/test_auth_error_middleware.py
+++ b/tests/test_auth_error_middleware.py
@@ -1,6 +1,7 @@
 """Tests for UpstreamAuthErrorMiddleware."""
 
 import json
+from http import HTTPStatus
 
 import pytest
 
@@ -14,7 +15,7 @@ from utils.error_handler import (
 # Default body that mimics http_client's 401 error response dict.
 _DEFAULT_401_BODY = {
     'error': 'HTTP Error',
-    'status_code': 401,
+    'status_code': HTTPStatus.UNAUTHORIZED,
     'message': 'Unauthorized',
     'mfa_required': False,
 }
@@ -51,7 +52,7 @@ class _MockApp:
         await send(
             {
                 'type': 'http.response.start',
-                'status': 200,
+                'status': HTTPStatus.OK,
                 'headers': [(b'content-type', b'application/json')],
             }
         )
@@ -123,7 +124,7 @@ async def test_no_flag_passes_through():
     mw = _make()
     sent = await _run(mw)
     status, _, body = await _collect_response(sent)
-    assert status == 200
+    assert status == HTTPStatus.OK
     assert 'ok' in body
 
 
@@ -137,7 +138,7 @@ async def test_flag_triggers_401():
     sent = await _run(mw)
     status, headers, body = await _collect_response(sent)
 
-    assert status == 401
+    assert status == HTTPStatus.UNAUTHORIZED
     assert 'www-authenticate' in headers
     assert 'invalid_token' in headers['www-authenticate']
     assert (
@@ -154,7 +155,7 @@ async def test_mfa_flag_includes_mfa_scope():
     sent = await _run(mw)
     status, headers, body = await _collect_response(sent)
 
-    assert status == 401
+    assert status == HTTPStatus.UNAUTHORIZED
     assert 'offline_access mfa' in headers['www-authenticate']
     assert 'MFA' in json.loads(body)['error_description']
 
@@ -166,7 +167,7 @@ async def test_non_mfa_flag_excludes_mfa_scope():
     sent = await _run(mw)
     status, headers, _ = await _collect_response(sent)
 
-    assert status == 401
+    assert status == HTTPStatus.UNAUTHORIZED
     assert 'mfa' not in headers['www-authenticate']
 
 
@@ -187,12 +188,12 @@ async def test_cooldown_passes_through_on_second_401():
 
     # First: 401
     sent1 = await _run(mw, scope=scope)
-    assert (await _collect_response(sent1))[0] == 401
+    assert (await _collect_response(sent1))[0] == HTTPStatus.UNAUTHORIZED
 
     # Second (same client, within cooldown): pass through as tool error
     sent2 = await _run(mw, scope=scope)
     status2, _, body2 = await _collect_response(sent2)
-    assert status2 == 200
+    assert status2 == HTTPStatus.OK
     assert 'status_code' in body2
 
 
@@ -216,12 +217,12 @@ async def test_per_client_cooldown_isolation():
 
     # Client A: 401
     sent_a = await _run(mw, scope=_http_scope(f'Bearer {token_a}'))
-    assert (await _collect_response(sent_a))[0] == 401
+    assert (await _collect_response(sent_a))[0] == HTTPStatus.UNAUTHORIZED
 
     # Client B (different token): swap app and test
     mw.app = app_b
     sent_b = await _run(mw, scope=_http_scope(f'Bearer {token_b}'))
-    assert (await _collect_response(sent_b))[0] == 401
+    assert (await _collect_response(sent_b))[0] == HTTPStatus.UNAUTHORIZED
 
 
 @pytest.mark.asyncio
@@ -260,7 +261,7 @@ async def test_stale_signal_triggers_401_on_next_request():
     status, headers, _ = await _collect_response(sent)
 
     # Signal is consumed → 401 returned to trigger re-auth
-    assert status == 401
+    assert status == HTTPStatus.UNAUTHORIZED
     assert 'www-authenticate' in headers
 
 
@@ -281,7 +282,7 @@ async def test_flag_triggers_401_with_any_body():
     sent = await _run(mw, scope=_http_scope(f'Bearer {token}'))
     status, headers, _ = await _collect_response(sent)
 
-    assert status == 401
+    assert status == HTTPStatus.UNAUTHORIZED
     assert 'www-authenticate' in headers
 
 
@@ -292,7 +293,11 @@ async def test_no_signal_passes_through_regardless_of_body():
 
     # Body looks like a 401 error but no signal was set
     app = _MockApp(
-        body={'error': 'HTTP Error', 'status_code': 401, 'message': 'Unauthorized'}
+        body={
+            'error': 'HTTP Error',
+            'status_code': HTTPStatus.UNAUTHORIZED,
+            'message': 'Unauthorized',
+        }
     )
     mw = UpstreamAuthErrorMiddleware(app)
 
@@ -300,7 +305,7 @@ async def test_no_signal_passes_through_regardless_of_body():
     status, _, body = await _collect_response(sent)
 
     # No signal → passes through as 200
-    assert status == 200
+    assert status == HTTPStatus.OK
     assert '401' in body
 
 
@@ -339,7 +344,7 @@ async def test_exception_triggers_401():
     sent = await _run(mw)
     status, headers, _ = await _collect_response(sent)
 
-    assert status == 401
+    assert status == HTTPStatus.UNAUTHORIZED
     assert 'www-authenticate' in headers
     assert 'mfa' in headers['www-authenticate']
 
@@ -353,7 +358,7 @@ async def test_exception_non_mfa_triggers_401_without_mfa_scope():
     sent = await _run(mw)
     status, headers, _ = await _collect_response(sent)
 
-    assert status == 401
+    assert status == HTTPStatus.UNAUTHORIZED
     assert 'www-authenticate' in headers
     assert 'mfa' not in headers['www-authenticate']
 
@@ -383,10 +388,10 @@ async def test_exception_respects_cooldown():
     # First call: should get 401
     sent1 = await _run(mw)
     status1, _, _ = await _collect_response(sent1)
-    assert status1 == 401
+    assert status1 == HTTPStatus.UNAUTHORIZED
 
     # Second call within cooldown: should get 500 fallback (no buffered response
     # because _RaisingApp raises before writing any response)
     sent2 = await _run(mw)
     status2, _, _ = await _collect_response(sent2)
-    assert status2 == 500
+    assert status2 == HTTPStatus.INTERNAL_SERVER_ERROR
--- a/utils/oauth.py
+++ b/utils/oauth.py
@@ -306,7 +306,9 @@ def _allow_browser_clients(handler):
     @wraps(handler)
     async def with_cors(request: Request) -> Response:
         if request.method == 'OPTIONS':
-            return Response(status_code=204, headers=_CORS_PREFLIGHT_HEADERS)
+            return Response(
+                status_code=HTTPStatus.NO_CONTENT, headers=_CORS_PREFLIGHT_HEADERS
+            )
 
         response = await handler(request)
         response.headers['Access-Control-Allow-Origin'] = '*'
@@ -505,7 +507,9 @@ def register_oauth_routes(mcp_server):
         try:
             config = _get_oauth_config()
         except ValueError as e:
-            return JSONResponse({'error': str(e)}, status_code=500)
+            return JSONResponse(
+                {'error': str(e)}, status_code=HTTPStatus.INTERNAL_SERVER_ERROR
+            )
 
         server_url = _get_server_url(request)
 
@@ -542,7 +546,9 @@ def register_oauth_routes(mcp_server):
         try:
             config = _get_oauth_config()
         except ValueError as e:
-            return JSONResponse({'error': str(e)}, status_code=500)
+            return JSONResponse(
+                {'error': str(e)}, status_code=HTTPStatus.INTERNAL_SERVER_ERROR
+            )
 
         # Forward all query parameters to Auth0
         params = dict(request.query_params)
@@ -700,7 +706,7 @@ def register_oauth_routes(mcp_server):
                 {'error': str(e)}, status_code=HTTPStatus.INTERNAL_SERVER_ERROR
             )
 
-        response = RedirectResponse(url=auth0_url, status_code=302)
+        response = RedirectResponse(url=auth0_url, status_code=HTTPStatus.FOUND)
         _set_nonce_cookie(response, nonce)
         return response
 
@@ -716,7 +722,9 @@ def register_oauth_routes(mcp_server):
         try:
             config = _get_oauth_config()
         except ValueError as e:
-            return JSONResponse({'error': str(e)}, status_code=500)
+            return JSONResponse(
+                {'error': str(e)}, status_code=HTTPStatus.INTERNAL_SERVER_ERROR
+            )
 
         # Parse request body
         body = await request.body()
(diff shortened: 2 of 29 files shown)


utils/recovery_hints.py
```diff
@@ -1,91 +1,81 @@
 """Recovery hints for error responses to help LLM self-recover."""
 
+from http import HTTPStatus
 from typing import Any
 
 # Hint registry: (status_code, domain) -> {recovery_hints, related_tools}
 _HINT_REGISTRY: dict[tuple[int, str], dict[str, list[str]]] = {
```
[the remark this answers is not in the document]
jisung-02: Not applying this — the current annotation is the correct one, and both suggested alternatives would be wrong here.

`HTTPStatus` subclasses `int` and `tuple` is covariant, so every `(HTTPStatus.X, 'domain')` key is a valid `tuple[int, str]`. Dict invariance does not apply: the literal is checked against the declared annotation rather than inferred separately and then assigned, so there is no `dict[tuple[HTTPStatus, str], ...]` -> `dict[tuple[int, str], ...]` conversion to reject. `mypy utils/recovery_hints.py` reports `Success: no issues found in 1 source file`.

The `int` key type is also load-bearing. Lookups pass a plain `int`: `_parse_status_code(raw: int | str | None) -> int | None` feeds `_HINT_REGISTRY.get((code, domain))` at lines 169 and 173. Narrowing the key type to `HTTPStatus` would break those call sites; widening it to `int | HTTPStatus` is a no-op because `HTTPStatus` is already a subtype of `int`; and casting the keys to `int` would undo the point of this PR.

utils/oauth.py
```diff
@@ -1119,7 +1129,7 @@ async def oauth_callback(request):
                     # MFA token is discarded — we only need the side effect
                     # of MFA completion in the Auth0 session. Log non-2xx
                     # responses for debugging misconfiguration.
-                    if mfa_response.status_code >= 400:
+                    if mfa_response.status_code >= HTTPStatus.BAD_REQUEST:
```
junho226, line 1132: Optional, and you already called this out in the PR description.

The condition now reads `>= HTTPStatus.BAD_REQUEST` directly beneath a comment saying "Log non-2xx responses", and the named constant makes the mismatch louder than the bare `400` did, since 3xx is skipped. Leaving a pre-existing inaccuracy alone is a defensible default, but this line is already being touched, so changing "non-2xx" to "error" would close it at zero risk. Your call either way.

tests/test_auth_error_middleware.py
```diff
@@ -137,7 +138,7 @@ async def test_flag_triggers_401():
     sent = await _run(mw)
     status, headers, body = await _collect_response(sent)
 
-    assert status == 401
+    assert status == HTTPStatus.UNAUTHORIZED
```
junho226, line 141: Suggestion, and it applies to the middleware tests generally rather than this line specifically.

These assertions now name the same symbol the production code sets, so the test no longer pins the wire value independently. If the middleware later changed to `HTTPStatus.FORBIDDEN`, a matching edit here would keep the suite green while the externally observable contract silently changed. A literal `401` would have caught that.

This matters more here than elsewhere in the sweep, because the ASGI `'status'` field is the wire contract itself. For assertions on internal dict plumbing the named constant genuinely reads better and I would leave those exactly as you have them. It is only worth considering for the handful of tests that assert the emitted HTTP status.

junho226 (dismissed):
## Summary

Mechanical refactor replacing HTTP status literals with `http.HTTPStatus` across 5 production modules and 23 test files, plus removal of the dead `format_user_friendly_error`. I verified the "no behavior change" claim independently rather than reading the diff, and it holds.

### Verification performed

**Token-level equivalence proof.** Rather than eyeballing 351 substitutions, I tokenized both the base (`cbd6809`) and head (`63c7a32`) versions of all 29 changed files, normalized every `HTTPStatus.MEMBER` back to its integer value on *both* sides, dropped comments and the added imports, then diffed the resulting token streams.

22 of 29 files come out token-for-token identical. The 7 that differ do so only in:

- trailing commas and `(`/`)` inserted by `ruff format` re-wrapping the now-longer lines (6 files), which are semantically inert
- the `format_user_friendly_error` and module-level `logger` deletion in `utils/error_handler.py` (1 file)

Every replacement is therefore value-preserving by construction, and no site maps a literal to the wrong member. Member histogram: `BAD_REQUEST` x90, `OK` x61, `NOT_FOUND` x55, `INTERNAL_SERVER_ERROR` x32, `FOUND` x30, `UNAUTHORIZED` x27, `FORBIDDEN` x26, `CREATED` x14, `NO_CONTENT` x9, `BAD_GATEWAY` x2, `TOO_MANY_REQUESTS` x2, `IM_A_TEAPOT` x2, `SERVICE_UNAVAILABLE` x1.

**The ASGI `status` field.** This is the one change where an `IntEnum` crosses into third-party code, so I measured it rather than relying on the equivalence proof. I ran a probe ASGI app sending `'status': HTTPStatus.UNAUTHORIZED` under both uvicorn HTTP implementations, against a plain-`401` control:

```
uvicorn 0.52.3 | h11 0.16.0 | httptools 0.8.0
h11:       b'HTTP/1.1 401 Unauthorized'
httptools: b'HTTP/1.1 401 Unauthorized'
```

Byte-identical to the `int` control on both paths. That is a newer stack than the uvicorn 0.34.3 / h11 0.16.0 you measured against, so the claim holds and extends.

**IntEnum vs int semantics on 3.12**, confirmed empirically: `str()`, f-string interpolation and `%s` all render `'404'`; `json.dumps` emits `404` as a value and `"404"` as a dict key; `==` and `hash()` match `int`. The only divergences are `repr()` and `type(x) is int`. I grepped for both and found no site that applies either to a status value (the four `repr`/`!r` hits are unrelated: log sanitization at `utils/oauth.py:159`, an MFA claim at `utils/security_settings.py:232`, and resource kwargs in `tools/resources.py`).

**Registry lookups.** `_HINT_REGISTRY` keys became `HTTPStatus` while `_parse_status_code` still returns a plain `int` through `int(raw)`, so lookups match on hash and equality. Keeping the annotation as `dict[tuple[int, str], ...]` is the correct choice here, not an oversight.

**Completeness, checked independently.** I walked every `NUMBER` token in all 82 `.py` files outside `.venv`, looking for values in 100-599, which is a wider net than the 2xx-5xx sweep described in the PR. 44 remain and none is a status code: `valid_days=365`, `page_size=100`, `hours > 168`, `timeout=300`, `_LOG_VALUE_MAX_CHARS=512`, `text[:200]`, `_CACHE_TTL=300`, and assorted mock ids and sizes. Nothing was missed.

**Dead-code removal.** `format_user_friendly_error` has zero references anywhere in the repo, including tests, docs and packaging metadata, and there is no `__all__` re-exporting it. It keyed off string codes (`'400'`, `'404'`), which is exactly what the issue flagged as needing a prerequisite decision, so deleting the table settles that question instead of deferring it. Isolating it in its own commit (`a679e51`) was the right way to do it.

### Checks run

| Check | Result |
|---|---|
| `ruff check .` | All checks passed |
| `ruff format --check .` | 96 files already formatted |
| `mypy --ignore-missing-imports --no-strict-optional .` | Success: no issues found in 82 source files |
| `pytest tests/ --cov=. --cov-fail-under=60` | 1127 passed, coverage 95.03% |

The 1127 count matches the PR description. It is also structurally guaranteed: no test function was added or removed anywhere in the sweep, which the token-level comparison confirms.

Everything above ran on Python 3.12.9 locally. CI additionally covers 3.13 and 3.14, which I did not run; nothing in this change is version-sensitive above 3.12, since the `IntEnum.__str__` change landed in 3.11.

### Notes

Two optional suggestions inline, neither affecting the verdict: the pre-existing "non-2xx" comment in `utils/oauth.py` that this change makes more visible, and a thought about named constants in the assertions that pin the emitted HTTP status.

Nice work on the verification write-up in the PR body. Every claim I checked was accurate, including the ones that would have been easy to hand-wave.

**Verdict: APPROVE.**

utils/oauth.py
jisung-02: Fixed in 867fd18 — changed "non-2xx" to "error" so the comment matches the `>= HTTPStatus.BAD_REQUEST` condition. Agreed that the named constant made the mismatch louder, and since the line was already being touched there was no reason to leave it.

tests/test_auth_error_middleware.py
jisung-02: Keeping the constants here. `HTTPStatus.UNAUTHORIZED` is a stdlib constant permanently equal to 401, not a reference to what the middleware sets — so if the middleware switched to `HTTPStatus.FORBIDDEN`, this assertion would compare 403 == 401 and fail. The assertion still pins the wire value independently.

The silent-drift scenario requires someone to edit the test to match, and a literal offers the same (non-)protection there: changing `401` to `403` alongside the production edit keeps the suite green just as quietly. The one real difference is friction against a mechanical rename sweep ("replace UNAUTHORIZED with FORBIDDEN") catching the test in the same pass, which felt too narrow to justify breaking this file out of the sweep's consistency.

junho226 (approved):
## Re-review: delta `63c7a32..867fd18`

`63c7a32` is an ancestor of `867fd18`, so nothing was rewritten. The verification behind the earlier approval—token-level equivalence of all 351 status-literal substitutions, ASGI `status` behavior under uvicorn/h11/httptools, `IntEnum`-vs-`int` semantics, `_HINT_REGISTRY` key and lookup compatibility, the leftover 100–599 literal sweep, and the `format_user_friendly_error` removal—stands unchanged for everything outside this delta. I did not redo it; below is what the new commit changed and what I re-verified at the current head.

### What changed

One commit, one line, comment-only:

- `867fd18 docs(oauth): align MFA logging comment with the >= 400 condition`—`utils/oauth.py:1130` now reads "Log error responses" instead of "Log non-2xx responses".

`git diff --stat` over the delta is 1 file, +1/-1, and the changed line is inside a `#` comment. No executable code, no guard, and no status constant moved.

### The two prior suggestions

1. **`utils/oauth.py` "non-2xx" comment—addressed, and correctly.** The guard at `utils/oauth.py:1132` is `>= HTTPStatus.BAD_REQUEST`, which admits 4xx and 5xx and skips 3xx. "Error responses" describes exactly that set, so the comment and the condition now agree.

2. **`tests/test_auth_error_middleware.py:141` named constants—declined, and I am not carrying it forward.** The rebuttal is right. That file imports `HTTPStatus` from the stdlib (`tests/test_auth_error_middleware.py:4`), not from `utils.auth_error_middleware`, so `HTTPStatus.UNAUTHORIZED` is permanently 401 and does not follow a production switch to `FORBIDDEN`—the assertion would compare 403 against 401 and fail. My original concern assumed a symbol shared with the middleware, which is not what the test does, so the assertion still pins the wire value independently. Keeping the file inside the sweep's consistency is the better trade.

### Re-verified at `867fd18`

- Delta is comment-only: no status literal reintroduced, no behavior or guard touched.
- Repo-wide sweep for bare 100–599 literals in comparison, assignment, and dict-value position (`status`/`status_code`): no hits outside `.venv`.
- No `non-2xx` or `non-200` wording left anywhere in the tree.
- Spot-checked the three surviving `Nxx` range comments against their guards, since this delta was precisely a comment/guard alignment fix: `tools/metrics_tools.py:371` (describes `unwrap_http_result`, no guard to drift from), `utils/http_client.py:257` (`except httpx.HTTPStatusError` genuinely covers 4xx and 5xx, and the `401` mention two lines below matches its `!= HTTPStatus.UNAUTHORIZED` guard), and `tools/work_session_tools.py:299` (describes the upstream 202 while the code branches on a body marker). All accurate.

### Checks run at `867fd18`

| Check | Result |
|---|---|
| `ruff check .` | pass—all checks passed |
| `ruff format . --check` | pass—82 files already formatted |
| `mypy --ignore-missing-imports --no-strict-optional .` | pass—no issues in 82 source files |
| `pytest tests/ --cov=. --cov-fail-under=60` | pass—1127 passed, total coverage 95.03% |

Nothing was skipped; every check in `.github/workflows/test.yml` ran locally, plus `ruff format --check`.

**Verdict: approve.**

Issue #166, opened 2026-08-06
build(deps): install ruff and pre-commit as dev dependencies, drop black, isort, and flake8


jisung-02:
The dev extra installs `black`, `isort`, and `flake8`, which nothing calls, and does not install `ruff` or `pre-commit`, which both `docs/installation.md` and `CONTRIBUTING.md` call and CI runs on every push. `[tool.ruff.lint]` already selects the rule sets the three cover.

#164 works around it in the documents with `uv tool install ruff` and `uv tool install pre-commit`, so a fresh checkout following the guide works today. Moving the two tools into the dev extra and dropping the three unused ones is a `pyproject.toml` and `uv.lock` change, and the `uv tool install` lines come back out of both documents once it lands.

Found while verifying the docs against the code in #164.

PR #184, opened 2026-08-18
ci: check formatting and pin the ruff version

### LINKED ISSUE #168: ci: add ruff format --check so Python blocks in markdown stay formatted
`ruff format` reaches into fenced Python inside markdown, but CI runs only `ruff check`, so nothing in the pipeline notices when a document drifts from the formatter.

#164 brings `CONTRIBUTING.md` and `docs/troubleshooting.md` in line because the pre-commit hook rewrote them on every commit. Any other document with Python blocks will do the same to whoever touches it next. Adding `ruff format --check` to `.github/workflows/test.yml` makes that visible in review instead of surprising in a local commit.

jisung-02:
## Background

The pre-commit config runs two ruff hooks, `ruff` and `ruff-format`, but `.github/workflows/test.yml` ran only `ruff check`. The two subcommands look at different things. The linter reads what the code means — unused imports, import order (`I`), security patterns (`S`), naming (`N`), bugbear (`B`). The formatter reads how it looks — line breaks, indentation, quote style, spacing. Neither substitutes for the other. `pyproject.toml:118` disables E501 with the comment `# line too long (formatter handles this)`, and since CI never ran the formatter, nothing in the pipeline was checking line length at all. Any commit that skipped the hooks carried its formatting through review untouched.

The ruff version was also unmanaged. The hook was pinned at `v0.9.7` while CI installed whatever `uv pip install ruff` resolved to on the day the job ran.

## Changes

`.github/workflows/test.yml` gains a `Check formatting with ruff` step running `ruff format --check .`, and the install in the lint step is pinned to `ruff==${{ env.RUFF_VERSION }}` with `RUFF_VERSION: "0.15.13"` declared on the job. Formatter output changes between releases, so without a pin a ruff release turns open PRs red without anyone touching the code.

`.pre-commit-config.yaml` moves the ruff rev from `v0.9.7` to `v0.15.13` to name the same release the workflow pins. The two pins are tied together only by the comment above `RUFF_VERSION`; nothing enforces it, and a mismatch surfaces as a CI failure on a tree that passed locally.

`CONTRIBUTING.md` and `docs/installation.md` list `ruff format --check` among the checks CI runs, and step 4 of the setup in `CONTRIBUTING.md` now runs it next to `ruff check` and `mypy`, so a contributor following the setup verbatim exercises every gate CI blocks on.

The format check is a separate step rather than a second line in the lint step so the Actions summary names which check failed.

## Markdown is deliberately out of scope

#168 was filed on the premise that `ruff format` reaches into fenced Python inside markdown and that the pre-commit hook rewrites `CONTRIBUTING.md` and `docs/troubleshooting.md` on every commit. That premise does not hold:

- The upstream hook declares `types_or: [python, pyi, jupyter]` (astral-sh/ruff-pre-commit `v0.15.13`, `.pre-commit-hooks.yaml`). It has never passed a markdown file to ruff.
- Ruff can format markdown, but only under `--preview`. Without it: `error: Failed to format README.md: Markdown formatting is experimental, enable preview mode.`
- That formatter reformats the document, not only its fenced Python — line breaks, list markers, tables.

Putting 14 markdown files under an experimental formatter means the documentation set gets rewritten whenever ruff changes it, and the pre-commit hook cannot check markdown, so the failure would appear only in CI with no local command to reproduce it. A markdown check is worth revisiting once the feature leaves preview and the hook covers it.

## Testing

The repository is already clean at the pinned version, so this PR reformats nothing:

```
$ ruff check .
All checks passed!

$ ruff format --check .
82 files already formatted
```

CI on this PR is the real check — it runs the new step across Python 3.12, 3.13, and 3.14.

## Follow-ups

`CONTRIBUTING.md:35` and `docs/installation.md:399` still say `uv tool install ruff` with no version, so a contributor can install a release whose formatter disagrees with the CI pin. That belongs with #166, which moves `ruff` and `pre-commit` into the dev extra.

Closes #168

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -20,6 +20,9 @@ jobs:
 
     env:
       UV_PYTHON: python${{ matrix.python-version }}
+      # Formatter output differs between releases; keep in step with the rev
+      # in .pre-commit-config.yaml.
+      RUFF_VERSION: "0.15.13"
 
     steps:
     - uses: actions/checkout@v6
@@ -45,8 +48,11 @@ jobs:
 
     - name: Lint with ruff
       run: |
-        uv pip install ruff
+        uv pip install ruff==${{ env.RUFF_VERSION }}
         uv run --no-sync ruff check .
+
+    - name: Check formatting with ruff
+      run: uv run --no-sync ruff format --check .
 
     - name: Type check with mypy
       run: |
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.9.7
+    rev: v0.15.13
     hooks:
       - id: ruff
         args: [--fix]
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -58,8 +58,9 @@ Thank you for your interest in contributing to the Alpacon MCP server! This guid
    # Check a token against the API (prompts for region and workspace)
    python main.py test
 
-   # Lint and type check
+   # Lint, formatting, and type check
    ruff check .
+   ruff format --check .
    mypy --ignore-missing-imports --no-strict-optional .
    ```
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -72,7 +72,7 @@ We use the following tools for code quality:
 - **ruff** for linting, import sorting, and formatting
 - **mypy** for type checking
 
-Both run as pre-commit hooks, and CI runs `ruff check`, `mypy`, and `pytest` under a coverage floor set in `.github/workflows/test.yml`.
+Both run as pre-commit hooks, and CI runs `ruff check`, `ruff format --check`, `mypy`, and `pytest` under a coverage floor set in `.github/workflows/test.yml`.
 
 ```bash
 # Format code
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -413,7 +413,7 @@ ruff format .
 mypy --ignore-missing-imports --no-strict-optional .
 ```
 
-CI runs the same three: `ruff check`, `mypy`, and `pytest` under a coverage floor. `.github/workflows/test.yml` holds the current threshold.
+CI runs the same checks: `ruff check`, `ruff format --check`, `mypy`, and `pytest` under a coverage floor. `.github/workflows/test.yml` holds the current threshold.
 
 ### Custom configuration
 


.github/workflows/test.yml
```diff
@@ -20,6 +20,9 @@ jobs:
 
     env:
       UV_PYTHON: python${{ matrix.python-version }}
+      # Formatter output differs between releases; keep in step with the rev
+      # in .pre-commit-config.yaml.
+      RUFF_VERSION: "0.15.13"
```
junho226, line 25: Verified this pin is coherent: `ruff==0.15.13` is on PyPI and `astral-sh/ruff-pre-commit` tags `v0.15.13`, so this and the hook rev name the same release. Nothing else in the repo pins or declares ruff, so there is no third value to fall out of step.

One thing to keep in mind for whoever moves this next: at 0.16.x ruff formats markdown by default, no `--preview` needed, so bumping past 0.15.x quietly widens what `ruff format --check .` covers from 82 files to 96. It is free today — the docs already pass under 0.16.3 — but it is a scope change hiding in a version bump, so it deserves its own commit rather than riding along with an unrelated one.

.github/workflows/test.yml
```diff
@@ -45,9 +48,12 @@ jobs:
 
     - name: Lint with ruff
       run: |
-        uv pip install ruff
+        uv pip install ruff==${{ env.RUFF_VERSION }}
         uv run --no-sync ruff check .
 
+    - name: Check formatting with ruff
+      run: uv run --no-sync ruff format --check .
```
junho226, line 55: Correct as written — this reuses the ruff that the preceding step installed into the venv, so no second install is needed.

Worth noting the rationale in the PR body only half pays off: the step is split so the Actions summary names which check failed, but since a failing `ruff check` aborts the job, a lint failure still hides the format result. You get the naming benefit when lint passes and format does not, which is the common case anyway. Not worth changing.

junho226 (dismissed):
Approving. The gate is real, the pin is coherent, and the tree it lands on is clean at that pin — I verified each of those rather than reading the diff for plausibility.

## What I ran

In a worktree at `bf61c36`, with the exact pinned version:

```
$ uvx ruff@0.15.13 --version
ruff 0.15.13
$ uvx ruff@0.15.13 check .
All checks passed!
$ uvx ruff@0.15.13 format --check .
82 files already formatted
```

Both exit 0, so the new step does not break CI on merge. I checked the same at `origin/main` (`82 files already formatted`, exit 0), so the gate is also clean on the target branch, not only on this branch. `actionlint .github/workflows/test.yml` exits 0, and a YAML parse confirms `RUFF_VERSION` resolves on the job env with the steps in the intended order. CI on the PR is green across 3.12, 3.13, and 3.14.

On the pin agreeing with the rest of the tooling: `ruff==0.15.13` exists on PyPI and `astral-sh/ruff-pre-commit` tags `v0.15.13`, so the workflow and `.pre-commit-config.yaml` name the same release rather than two that merely look alike. Nothing else in the repo runs or declares ruff — `docker.yml` and `publish.yml` do not invoke it, and neither `[project.optional-dependencies]` nor `[dependency-groups]` declares it — so there is no third pin left behind to disagree with. The pin is applied everywhere ruff actually runs in CI.

## Merge order against #183

No hazard in either direction, verified rather than assumed. I extracted `origin/143-httpstatus-magic-numbers` to a scratch tree and ran the pinned ruff over it: `82 files already formatted`, exit 0, and `All checks passed!`. So if #184 lands first, #183 passes the new gate once it picks up main; if #183 lands first, this branch still passes. The two touch disjoint files.

## Notes, none blocking

**The markdown rationale has a shelf life tied to the pin.** The PR body argues markdown is out of scope because ruff only formats it under `--preview`. That is exactly right at the pinned version:

```
$ uvx ruff@0.15.13 format --check CONTRIBUTING.md
error: Failed to format CONTRIBUTING.md: Markdown formatting is experimental, enable preview mode.
```

It stops being right at 0.16.x, where markdown is formatted by default with no flag:

```
$ uvx ruff@0.16.3 format --check CONTRIBUTING.md
1 file already formatted
$ uvx ruff@0.16.3 format --check .
96 files already formatted
```

That 96 is the same 82 plus the 14 markdown files, picked up through normal file discovery. Worth knowing when the pin next moves, because it changes what `ruff format --check .` covers without anyone editing the workflow. The good news is that it costs nothing today: the docs are already clean under 0.16.3, so whoever bumps the pin inherits a passing gate rather than a reformat commit. That also settles #168's original premise on its own terms — the concern was real, just early.

**The local install path is still unpinned.** `CONTRIBUTING.md:35` and `docs/installation.md:399` say `uv tool install ruff` with no version, so a contributor who set up recently is on 0.16.x while CI gates at 0.15.13. Right now that is harmless because the tree passes under both, but that is a coincidence of this tree, not a property of the setup. The PR body already calls this out and points at #166, which is open and scoped to move ruff and pre-commit into the dev extra — the right place for it, so I would not expand this PR to cover it.

**Small doc gap in the same file the PR touches.** `CONTRIBUTING.md:75` now lists `ruff format --check` among the CI checks, but the "Verify Setup" snippet at `CONTRIBUTING.md:62` still runs only `ruff check .` and `mypy`. A contributor following step 4 verbatim still does not run the thing CI now blocks on. One line in that block would close it, though it is equally reasonable to leave it for #166.

One thing I looked for and did not find a problem with: I expected the comment-only coupling between `RUFF_VERSION` and the pre-commit `rev` to be a drift risk from automated bumps, but there is no `.github/dependabot.yml` in this repo at all. Nothing will move either pin unilaterally, so the two only drift if a human edits one and forgets the other — which is what the comment is there to catch.

## What I could not verify

I ran ruff through `uvx` on macOS rather than reproducing the workflow's `uv pip install ruff==... && uv run --no-sync` sequence on a Linux runner. The formatter is deterministic across platforms for a given version, and the PR's own green CI already exercises the real path on all three Python versions, so I do not think anything is hiding there.

jisung-02:
Thanks for verifying the gate against both branches and #183 rather than reading the diff — that is the part I could not check from here.

Taking the third note, leaving the other two.

**Setup verification gap — fixed in 6d6d966.** `CONTRIBUTING.md` step 4 now runs `ruff format --check .` between `ruff check .` and `mypy`, so following the setup verbatim exercises every check CI blocks on. I left the two `ruff format .` lines further down alone: those are the write form in the day-to-day workflow block, and turning them into a check there would make the docs stop telling a contributor how to actually format their code. The PR body's Changes section names the edit now.

**Markdown scope at 0.16.x — noted, not acted on.** Agreed that the 82-to-96 widening is a scope change hiding in a version bump. It is not something this PR can encode; the pin comment does not carry it, and I would rather it live in the bump commit that causes it than as a comment that goes stale here.

**Unpinned local install — staying with #166.** Same read as yours. The tree passing under both 0.15.13 and 0.16.3 is a property of this tree, not the setup, but the fix is moving ruff and pre-commit into the dev extra, and that is #166's scope.

On the split-step rationale only half paying off: correct, and I will leave it. The naming benefit lands in the case where lint passes and format does not, which is the case the step exists for.

junho226 (approved):
Re-review of the delta only (`bf61c36..6d6d966`). The pin-coherence verification from the previous review stands for the unchanged portion; nothing outside `CONTRIBUTING.md` moved.

## What changed

One commit, `docs: verify formatting in the setup steps`, touching two lines of the **Verify Setup** snippet in `CONTRIBUTING.md`: the block comment becomes `# Lint, formatting, and type check`, and `ruff format --check .` is inserted between `ruff check .` and the mypy invocation. This closes the third note from the previous round — the snippet a contributor runs now includes the check CI newly blocks on.

## The fix is correct and complete

- **Spelling matches CI exactly.** The snippet reads `ruff format --check .`, byte-for-byte what `.github/workflows/test.yml:55` runs. No `ruff format . --check` variant anywhere.
- **Order matches CI.** Workflow runs `ruff check .` (line 52) then `ruff format --check .` (line 55) then mypy; the snippet keeps the same relative order.
- **The prose at `CONTRIBUTING.md:76` and the snippet now agree** — both name `ruff check`, `ruff format --check`, `mypy`, `pytest`.
- **No other setup path is left stale.** I swept every ruff mention in the repo. `docs/installation.md:405-413` and the code-style block at `CONTRIBUTING.md:78-90` invoke the *writer* (`ruff format .`), not `--check` — that is the local fix-it flow, and running it leaves the tree in a state where CI's `--check` passes, so neither is inconsistent with the new gate. `README.md` carries no lint or format instructions, and the repo has no `Makefile` or other task runner, so there is no third place to update.
- **No claim in the new text is false at the pinned version.** `ruff format --check .` is a valid 0.15.13 invocation and it passes at this head (82 files already formatted).

## One follow-up, non-blocking and out of scope

Documenting `ruff format --check .` as a local step raises the stakes of the still-unpinned local install (`uv tool install ruff` at `CONTRIBUTING.md:35` and `docs/installation.md:399`): a contributor on a newer ruff now runs the one command whose output is version-sensitive, so they can see a formatting verdict CI does not produce. That is issue #166's territory, correctly not this PR's.

## Verified at this head

| Check | Result |
|---|---|
| `uvx ruff@0.15.13 check .` | pass — all checks passed |
| `uvx ruff@0.15.13 format --check .` | pass — 82 files already formatted |
| `actionlint .github/workflows/test.yml` | pass — no findings |
| `gh pr checks 184` | all 6 green (test 3.12/3.13/3.14, CodeQL, Analyze actions/python) |

Nothing was left unrun.

PR #185, opened 2026-08-18
build: reconcile the two dev dependency lists in pyproject.toml

### LINKED ISSUE #170: build: reconcile the two dev dependency lists in pyproject.toml
`pyproject.toml` carries two dev dependency lists — `[project.optional-dependencies] dev` and `[dependency-groups] dev` — with `pytest` in both at different lower bounds.

Which one wins depends on how the environment is created (`uv pip install -e .[dev]` versus `uv sync`), so a fresh checkout can resolve a different pytest than CI. Reconciling them changes what gets installed, so it needs a decision about which list is authoritative rather than an edit.

jisung-02:
Closes #170.

## What was wrong

`pyproject.toml` carried two dev dependency lists that disagreed:

| | `[project.optional-dependencies] dev` | `[dependency-groups] dev` |
|---|---|---|
| pytest | `>=7.0.0` | `>=8.4.2` |
| pytest-asyncio | `>=0.21.0` | `>=1.2.0` |
| pytest-cov | absent | `>=6.0.0` |

Which one applied depended on how the environment was created: `uv sync` installs the uv-only group, `uv pip install -e .[dev]` installs the extra and ignores the group. CI used the latter, so it never saw pytest-cov and installed it as a separate step.

## What this changes

**One authoritative list.** The PEP 621 extra wins—it is the list that survives outside uv. `[dependency-groups]` is gone and the pytest floors move up to what the lockfile already resolved. uv synced that group by default, so one behavior does change, and it is sharper than "stops installing": in a development checkout a bare `uv sync` now **uninstalls** the test dependencies. No documented bare-`uv sync` flow runs the tests, so nothing breaks in place. The gap is the other direction—a contributor who follows the setup block and later reaches for "Update installation" or the venv rebuild under "Virtual environment issues" loses pytest and mypy, and uv names the packages it removed without naming the extra that puts them back. Both recipes now carry `--extra dev` as a note, and the two setup blocks say the flag belongs on every later sync. The note stays a note because both recipes serve runtime installs too.

**black, isort, and flake8 are dropped.** ruff and pre-commit took over the role in f9fe2bc and 5369d95 rewrote the docs to `ruff check` / `ruff format`; the `pyproject.toml` entries were the only leftover, and no workflow ever ran them. `uv.lock` loses 146 lines. The ruff selectors named `flake8-bugbear` and `isort` are built into ruff and are unaffected.

**CI installs from the lockfile.** `uv sync --extra dev --locked` installs exactly what `uv.lock` pins and fails when the lockfile and `pyproject.toml` drift. mypy comes from the dev extra now, so its ad-hoc install is gone, and the `uv venv` step goes with it—`uv sync` creates the environment under `UV_PYTHON` by itself.

**Docs point at a flow that installs the dev tooling.** `uv pip install -e .[dev]` becomes `uv sync --extra dev`. README's "Option C: development installation" only ever ran `uv sync`, so it is renamed to "Option C: run from a checkout".

**Coverage artifacts are ignored.** Neither `.coverage` nor `htmlcov/` was in `.gitignore`, though `CONTRIBUTING.md` documents `--cov-report=html`.

## Version pinning, deliberately asymmetric

`RUFF_VERSION` stays an exact pin while mypy stays a floor. `ruff format --check` blocks the merge, so a formatter release that changes output would break CI at random. The mypy step is `continue-on-error: true`, so drift costs nothing—and an exact pin in a published package's dev extra would leak onto anyone installing `.[dev]`. The comment on the mypy floor now says so.

The setup instructions held a third copy of the ruff version and were the only unpinned one, which left a local `ruff format --check` free to disagree with the gate. They now install `ruff==0.15.13` and name the rev that governs it, so `.github/workflows/test.yml`, `.pre-commit-config.yaml`, and the docs move together.

## Verification

The uninstall the docs now warn about, reproduced on this branch:

```
$ uv sync
Uninstalled 12 packages in 176ms
 - pytest==9.0.3
 - pytest-asyncio==1.3.0
 - pytest-cov==7.0.0
 ...

$ uv run --no-sync pytest --version
error: Failed to spawn: `pytest`
```

```
$ uv sync --extra dev --locked --dry-run
Resolved 48 packages in 18ms
Checked 42 packages in 7ms
Would make no changes

$ uv run --no-sync pytest tests/ -q --cov=. --cov-fail-under=60
Required test coverage of 60% reached. Total coverage: 95.00%
1127 passed, 7 warnings in 8.36s

$ uv run --no-sync ruff check .
All checks passed!

$ uv run --no-sync ruff format --check .
82 files already formatted
```

Dropping the `uv venv` step was the one change only CI could confirm, and this run does: "Install dependencies" succeeds on 3.12, 3.13, and 3.14 with no environment created ahead of it.

```
test (3.12): pass          Analyze (actions): pass
test (3.13): pass          Analyze (python): pass
test (3.14): pass          CodeQL: pass
```

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -29,7 +29,7 @@ Thank you for your interest in contributing to the Alpacon MCP server! This guid
    source .venv/bin/activate
 
    # Install development dependencies
-   uv pip install -e .[dev]
+   uv sync --extra dev
 
    # ruff and pre-commit ship as tools, not as dev dependencies
    uv tool install ruff
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -28,7 +28,8 @@ Thank you for your interest in contributing to the Alpacon MCP server! This guid
    uv venv
    source .venv/bin/activate
 
-   # Install development dependencies
+   # Install development dependencies. Keep --extra dev on every later
+   # uv sync too: a bare one uninstalls them.
    uv sync --extra dev
 
    # ruff and pre-commit ship as tools, not as dev dependencies.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -31,8 +31,9 @@ Thank you for your interest in contributing to the Alpacon MCP server! This guid
    # Install development dependencies
    uv sync --extra dev
 
-   # ruff and pre-commit ship as tools, not as dev dependencies
-   uv tool install ruff
+   # ruff and pre-commit ship as tools, not as dev dependencies.
+   # Keep the ruff pin in step with the rev in .pre-commit-config.yaml.
+   uv tool install ruff==0.15.13
    uv tool install pre-commit
 
    # Install pre-commit hooks
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -38,13 +38,8 @@ jobs:
     - name: Install uv
       uses: astral-sh/setup-uv@v8.1.0
 
-    - name: Create virtual environment
-      run: uv venv
-
     - name: Install dependencies
-      run: |
-        uv pip install -e .[dev]
-        uv pip install pytest-cov
+      run: uv sync --extra dev --locked
 
     - name: Lint with ruff
       run: |
@@ -55,9 +50,7 @@ jobs:
       run: uv run --no-sync ruff format --check .
 
     - name: Type check with mypy
-      run: |
-        uv pip install mypy
-        uv run --no-sync mypy --ignore-missing-imports --no-strict-optional .
+      run: uv run --no-sync mypy --ignore-missing-imports --no-strict-optional .
       continue-on-error: true
 
     - name: Test with pytest
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,11 @@ wheels/
 *.egg-info
 *__pycache__*
 
+# Coverage
+.coverage
+.coverage.*
+htmlcov/
+
 # Virtual environments
 .venv
 
--- a/README.md
+++ b/README.md
@@ -298,7 +298,7 @@ Or via environment variable (wins over the config file):
 export ALPACON_MCP_US1_PRODUCTION_URL="https://production.us1.alpacon.io"
 ```
 
-### Option C: development installation
+### Option C: run from a checkout
 ```bash
 git clone https://github.com/alpacax/alpacon-mcp.git
 cd alpacon-mcp
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -395,8 +395,9 @@ uv venv
 source .venv/bin/activate
 uv sync --extra dev
 
-# ruff and pre-commit ship as tools, not as dev dependencies
-uv tool install ruff
+# ruff and pre-commit ship as tools, not as dev dependencies.
+# Keep the ruff pin in step with the rev in .pre-commit-config.yaml.
+uv tool install ruff==0.15.13
 uv tool install pre-commit
 
 # Install pre-commit hooks (ruff + mypy)
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -393,7 +393,7 @@ cd alpacon-mcp
 # Install project dependencies
 uv venv
 source .venv/bin/activate
-uv pip install -e .[dev]
+uv sync --extra dev
 
 # ruff and pre-commit ship as tools, not as dev dependencies
 uv tool install ruff
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -390,7 +390,8 @@ sudo chown -R alpacon-mcp:alpacon-mcp /opt/alpacon-mcp
 git clone https://github.com/alpacax/alpacon-mcp.git
 cd alpacon-mcp
 
-# Install project dependencies
+# Install project dependencies. Keep --extra dev on every later uv sync
+# too: a bare one uninstalls them.
 uv venv
 source .venv/bin/activate
 uv sync --extra dev
@@ -461,7 +462,7 @@ echo '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"clientInfo":{"nam
 git pull origin main
 
 # Update dependencies to match the lockfile
-uv sync
+uv sync  # add --extra dev in a development checkout
 
 # Restart service if running
 sudo systemctl restart alpacon-mcp  # Linux service
@@ -518,7 +519,7 @@ chmod +x ~/.local/bin/uv
 rm -rf .venv
 uv venv
 source .venv/bin/activate
-uv sync
+uv sync  # add --extra dev in a development checkout
 ```
 
 ### Platform-specific issues
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -60,12 +60,12 @@ alpacon-mcp-http = "main_http:main"
 
 [project.optional-dependencies]
 dev = [
-    "pytest>=7.0.0",
-    "pytest-asyncio>=0.21.0",
-    "black>=23.0.0",
-    "isort>=5.12.0",
-    "flake8>=6.0.0",
-    "mypy>=1.0.0",
+    "pytest>=8.4.2",
+    "pytest-asyncio>=1.2.0",
+    "pytest-cov>=6.0.0",
+    # Floor only: the mypy step is advisory (continue-on-error), so CI need
+    # not match the pre-commit rev exactly.
+    "mypy>=1.15.0",
 ]
 
 [tool.hatch.version]
@@ -82,13 +82,6 @@ include = [
     "/tools/**/*.py",
     "/utils/**/*.py",
     "/config/token.json.example",
-]
-
-[dependency-groups]
-dev = [
-    "pytest>=8.4.2",
-    "pytest-asyncio>=1.2.0",
-    "pytest-cov>=6.0.0",
 ]
 
 [tool.ruff]
--- a/uv.lock
+++ b/uv.lock
@@ -19,16 +19,7 @@ dependencies = [
 
 [package.optional-dependencies]
 dev = [
-    { name = "black" },
-    { name = "flake8" },
-    { name = "isort" },
     { name = "mypy" },
-    { name = "pytest" },
-    { name = "pytest-asyncio" },
-]
-
-[package.dev-dependencies]
-dev = [
     { name = "pytest" },
     { name = "pytest-asyncio" },
     { name = "pytest-cov" },
@@ -36,24 +27,15 @@ dev = [
 
 [package.metadata]
 requires-dist = [
-    { name = "black", marker = "extra == 'dev'", specifier = ">=23.0.0" },
-    { name = "flake8", marker = "extra == 'dev'", specifier = ">=6.0.0" },
     { name = "httpx", specifier = ">=0.27.1" },
-    { name = "isort", marker = "extra == 'dev'", specifier = ">=5.12.0" },
     { name = "mcp", specifier = ">=1.28.1,<2" },
-    { name = "mypy", marker = "extra == 'dev'", specifier = ">=1.0.0" },
+    { name = "mypy", marker = "extra == 'dev'", specifier = ">=1.15.0" },
     { name = "pyjwt", extras = ["crypto"], specifier = ">=2.10.1" },
-    { name = "pytest", marker = "extra == 'dev'", specifier = ">=7.0.0" },
-    { name = "pytest-asyncio", marker = "extra == 'dev'", specifier = ">=0.21.0" },
+    { name = "pytest", marker = "extra == 'dev'", specifier = ">=8.4.2" },
+    { name = "pytest-asyncio", marker = "extra == 'dev'", specifier = ">=1.2.0" },
+    { name = "pytest-cov", marker = "extra == 'dev'", specifier = ">=6.0.0" },
 ]
 provides-extras = ["dev"]
-
-[package.metadata.requires-dev]
-dev = [
-    { name = "pytest", specifier = ">=8.4.2" },
-    { name = "pytest-asyncio", specifier = ">=1.2.0" },
-    { name = "pytest-cov", specifier = ">=6.0.0" },
-]
 
 [[package]]
 name = "annotated-types"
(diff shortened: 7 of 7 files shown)


CONTRIBUTING.md
```diff
@@ -29,10 +29,11 @@ Thank you for your interest in contributing to the Alpacon MCP server! This guid
    source .venv/bin/activate
 
    # Install development dependencies
-   uv pip install -e .[dev]
+   uv sync --extra dev
```
junho226, line 33: Non-blocking, and the docs may be fine as they are — but the behavior change is a step sharper than the description says.

Removing the default `dev` group turns bare `uv sync` from "does not install the test dependencies" into "uninstalls them". Verified on this branch, immediately after the command on this line:

```
$ uv sync
Resolved 48 packages in 3ms
Uninstalled 12 packages in 173ms
 - coverage==7.13.4
 - mypy==1.18.2
 - pytest==9.0.3
 - pytest-asyncio==1.3.0
 - pytest-cov==7.0.0
 - ruff==0.15.13
 ... (iniconfig, mypy-extensions, packaging, pathspec, pluggy, pygments)

$ uv run --no-sync pytest --version
error: Failed to spawn: `pytest`
```

On `main` the old lockfile carried `[package.dev-dependencies] dev = [pytest, pytest-asyncio, pytest-cov]`, and uv syncs the default group, so a bare `uv sync` there kept the test runner installed.

The claim that no documented bare-`uv sync` flow runs the tests holds — I checked all nine sites. The gap is the other direction: a contributor who sets up per this block and later follows `docs/installation.md` L521 ("Virtual environment issues", rebuild the venv then `uv sync`) or L464 ("Update installation") silently loses pytest and mypy, with nothing in the output naming the dev extra as the thing to re-add. The venv-rebuild recipe is the likelier trap, since reaching for it means something is already broken.

Cheapest fix is one word: make the L521 rebuild `uv sync --extra dev`, or note here that a later bare `uv sync` undoes this step.

junho226 (dismissed):
## Review: #185 — build: reconcile the two dev dependency lists in pyproject.toml

**Stack**: python (uv + hatchling) · **Changes**: +26 -171 across 7 files

### Summary

Collapses the two disagreeing dev dependency lists onto the PEP 621 `dev` extra, drops the dead black/isort/flake8 entries, and points CI at the lockfile with `uv sync --extra dev --locked`. Every claim in the description that I could reproduce locally reproduces, and the reasoning holds up under checking. One non-blocking note, inline.

### Before/after dependency sets

Compared the two lists explicitly rather than reading the diff:

| Package | `[optional-dependencies] dev` (before) | `[dependency-groups] dev` (before) | after |
|---|---|---|---|
| pytest | `>=7.0.0` | `>=8.4.2` | `>=8.4.2` |
| pytest-asyncio | `>=0.21.0` | `>=1.2.0` | `>=1.2.0` |
| pytest-cov | absent | `>=6.0.0` | `>=6.0.0` |
| mypy | `>=1.0.0` | absent | `>=1.15.0` |
| black | `>=23.0.0` | absent | dropped |
| isort | `>=5.12.0` | absent | dropped |
| flake8 | `>=6.0.0` | absent | dropped |

Nothing was silently dropped and no floor was loosened: every survivor keeps the higher of the two floors, and the mypy floor rises to meet the `mirrors-mypy` rev instead of sitting three years behind it.

The three removals are genuinely dead code. There is no `[tool.black]` and no `[tool.isort]` section, no `.flake8` / `setup.cfg` / `tox.ini` anywhere in the tree, and `git grep -E 'black|isort|flake8' -- .github/` returns nothing. `[tool.ruff.lint.isort]`, `[tool.ruff.lint.mccabe]`, and the `B` / `S` selectors are ruff's own implementations, so they are unaffected, as the description says.

`uv.lock` loses exactly those three plus their transitive closure — `mccabe`, `platformdirs`, `pycodestyle`, `pyflakes`, `pytokens` — and adds nothing. `platformdirs` is now absent from the file altogether, so no surviving package still needed it, and `coverage` is still present for pytest-cov. `[package.dev-dependencies]` is gone from the project's own block; `provides-extras = ["dev"]` stays.

### Tooling entry points still resolve

- `.github/workflows/test.yml` is the only thing in the repo that installs the dev list. `publish.yml` builds with `python -m build`, `docker.yml` never installs Python deps, and the `Dockerfile` installs a prebuilt wheel — none of them read the extra or the group. No `Makefile`, `tox.ini`, or `noxfile.py` exists.
- The `uv pip install ruff==${RUFF_VERSION}` step lands in the venv that `uv sync` just created, and every later step passes `--no-sync`, so nothing prunes ruff mid-run. Dropping the `uv venv` step was the one change only CI could confirm, and all three matrix legs are green.
- Every bare `uv sync` left in the docs is a runtime or repair flow, as claimed — README Option C, `docs/README.md`, `getting-started.md`, `installation.md` (eight sites), `troubleshooting.md`. None of them runs the tests.

### Checks run locally (uv 0.11.21, CPython 3.12.9)

```
$ uv lock --check
Resolved 48 packages in 3ms                          # exit 0

$ uv sync --extra dev --locked
... mypy==1.18.2  pytest==9.0.3  pytest-asyncio==1.3.0  pytest-cov==7.0.0

$ uv run --no-sync ruff check .
All checks passed!

$ uv run --no-sync ruff format --check .
82 files already formatted

$ uv run --no-sync mypy --ignore-missing-imports --no-strict-optional .
Success: no issues found in 82 source files

$ PYTHONPATH=. uv run --no-sync pytest tests/ --tb=short --cov=. --cov-report=term-missing --cov-fail-under=60
Required test coverage of 60% reached. Total coverage: 95.03%
1127 passed, 7 warnings in 12.84s
```

Ran ruff at the pinned `0.15.13`, matching the CI gate. The `.gitignore` addition works — the `.coverage` that pytest run produced is ignored (`git check-ignore -v .coverage` resolves to `.gitignore:11`).

### Highlights

- `--locked` converts lockfile drift into a build failure and retires two ad-hoc installs in the same stroke. That is the part of this PR with lasting value.
- The asymmetric pinning argument is right, and the `pyproject.toml` comment is where a future mypy bump will actually read it. An exact pin in a published package's dev extra really would leak onto anyone installing `.[dev]`.
- Copilot's two suppressed comments both asked for the ruff pin in the setup instructions, which 8d2ecfc already does. The literal now lives in four places (`test.yml`, `.pre-commit-config.yaml`, `CONTRIBUTING.md`, `docs/installation.md`) held together only by comments, but the description reasons about that deliberately and pre-commit still runs ruff at its own pinned rev, so the docs copy only governs ad-hoc CLI use.

**Verdict**: approve.

CONTRIBUTING.md
jisung-02: Right, and the framing correction matters more than the fix — "does not install" and "uninstalls" are different failure modes, and the description only claimed the first. Reproduced your run exactly: bare `uv sync` removes 12 packages and `uv run --no-sync pytest --version` then fails with `Failed to spawn: pytest`.

Fixed in 53b21b6. I took both sites rather than just L521, since "Update installation" is the same trap with a slower fuse:

- `docs/installation.md` L464 (Update installation) and L521 (venv rebuild): `uv sync  # add --extra dev in a development checkout`
- `CONTRIBUTING.md` L31 and `docs/installation.md` L393 setup blocks: a line saying `--extra dev` belongs on every later sync, not only the first one

Kept the flag as a trailing note instead of making it the default in those two recipes — both also serve the runtime install paths above them, where pulling in pytest and mypy would be wrong. The setup-block note is the part that generalizes: it covers the two recipes I patched plus any bare `uv sync` a contributor reaches for that I did not.

PR description updated to state the uninstall directly, with the repro.

junho226 (approved):
Re-review at `53b21b6`. My previous approval was at `8d2ecfc7` and was dismissed by this push.

**The documentation gap from that round is fixed.** `53b21b6` annotates both recipes I named and both setup blocks, and I re-enumerated every remaining bare `uv sync` in the repo to confirm the coverage is complete. Everything else re-verifies clean: the dependency sets, the lockfile, the tooling entry points, and the local checks.

### The prior finding is closed

`53b21b6` touches only the two docs, +6/-4:

| Site | Section | Now reads |
|---|---|---|
| `CONTRIBUTING.md:31-33` | Development setup | note + `uv sync --extra dev` |
| `docs/installation.md:393-397` | For contributing | note + `uv sync --extra dev` |
| `docs/installation.md:465` | Update installation | `uv sync  # add --extra dev in a development checkout` |
| `docs/installation.md:522` | Virtual environment issues | `uv sync  # add --extra dev in a development checkout` |

L465 and L522 are the exact two recipes from the finding, and the note at both setup blocks says the flag belongs on every later sync, which covers the "a later bare sync undoes this" direction head-on. I re-ran the underlying behavior at this HEAD, so the note is accurate rather than merely plausible:

```
$ uv sync
Resolved 48 packages in 3ms
Uninstalled 12 packages in 131ms
 - mypy==1.18.2
 - pytest==9.0.3
 - pytest-asyncio==1.3.0
 - pytest-cov==7.0.0
 - ruff==0.15.13
 ... (coverage, iniconfig, mypy-extensions, packaging, pathspec, pluggy, pygments)

$ uv run --no-sync pytest --version
error: Failed to spawn: `pytest`
```

### Coverage of the remaining bare-`uv sync` sites

Re-enumerated rather than carried over from last round. 18 `uv sync` lines in tracked files: 4 now carry the dev extra or its note (table above), `CLAUDE.md:30` already had `--extra dev`, `.github/workflows/test.yml:42` is `--extra dev --locked`, 2 are the comment lines themselves, and 10 stay bare:

| Site | Section | Flow |
|---|---|---|
| `README.md:306` | Option C: run from a checkout | runtime, ends in `python main.py` |
| `docs/README.md:30` | "To run from a checkout instead" | runtime |
| `docs/installation.md:79, 114, 136, 177, 237, 268` | Manual installation / Homebrew / system Python / Setup Alpacon MCP / Setup process / PowerShell | platform runtime installs |
| `docs/troubleshooting.md:51` | Server won't start (`ModuleNotFoundError`) | runtime repair |
| `docs/getting-started.md:82` | Method 3: Development setup | see the note at the end |

`pytest` and `mypy` appear after exactly two install blocks in the docs, `CONTRIBUTING.md:58/66` and `docs/installation.md:408/415`, and both install with `--extra dev`. No documented flow that runs the tests reaches for a bare sync.

### Dependency sets, before and after

| Package | `[optional-dependencies] dev` before | `[dependency-groups] dev` before | after |
|---|---|---|---|
| pytest | `>=7.0.0` | `>=8.4.2` | `>=8.4.2` |
| pytest-asyncio | `>=0.21.0` | `>=1.2.0` | `>=1.2.0` |
| pytest-cov | absent | `>=6.0.0` | `>=6.0.0` |
| mypy | `>=1.0.0` | absent | `>=1.15.0` |
| black | `>=23.0.0` | absent | dropped |
| isort | `>=5.12.0` | absent | dropped |
| flake8 | `>=6.0.0` | absent | dropped |

Every specifier that changed sits under the dev extra, and each change either raises a floor or adds a package. No runtime dependency moved and nothing was loosened, confirmed against the lockfile's `requires-dist` rather than by reading the diff.

Lockfile package set: 52 names to 44. The 8 that leave are the three dropped tools plus transitives that only they pulled (`mccabe`, `platformdirs`, `pycodestyle`, `pyflakes`, `pytokens`); `[tool.ruff.lint.mccabe]` is ruff's own built-in and is unaffected. Nothing was added: `pytest-cov` was already locked and only moves from `[package.dev-dependencies]` to `[package.optional-dependencies]`. Locked versions all clear the new floors (pytest 9.0.3, pytest-asyncio 1.3.0, pytest-cov 7.0.0, mypy 1.18.2).

### Tooling entry points

- `.github/workflows/test.yml` is the only consumer of the dev extra: `uv sync --extra dev --locked` is a blocking step, and every later step passes `--no-sync`, so nothing prunes the ruff pinned in at L46. CI does exercise the reconciled list.
- `Dockerfile` builds and installs a wheel (`pip wheel --no-deps` then `pip install`), `publish.yml` runs `python -m build` and `twine`, `docker.yml` installs no dependencies. None of them reads the extra or the removed group. No `Makefile`, `tox.ini`, or `noxfile.py` exists.
- No `black`, `isort`, or `flake8` reference survives outside the ruff selector names in `pyproject.toml`.

### Checks run locally (uv 0.11.21, macOS)

```
$ uv lock --check
Resolved 48 packages in 6ms                     # exit 0

$ uv sync --extra dev --locked
Resolved 48 packages in 3ms
Checked 42 packages in 13ms

$ uv run --no-sync ruff check .                 # ruff 0.15.13, the CI pin
All checks passed!

$ uv run --no-sync ruff format --check .
82 files already formatted

$ uv run --no-sync mypy --ignore-missing-imports --no-strict-optional .
Success: no issues found in 82 source files

$ PYTHONPATH=. uv run --no-sync pytest tests/ --tb=short --cov=. --cov-report=term-missing --cov-fail-under=60
Required test coverage of 60% reached. Total coverage: 95.03%
1127 passed, 7 warnings in 9.28s
```

Unverified locally: the 3.13 and 3.14 legs, and the dropped `uv venv` step, which only CI can confirm. All three matrix legs plus CodeQL are green at this HEAD.

### One non-blocking note, and it is not the old one

`docs/getting-started.md:82` is headed "Method 3: Development setup" and installs with a bare `uv sync`. Before this PR that block did hand you a test runner, because uv synced the default group; now it hands you none. Nothing in that file runs the tests, so no documented flow breaks. But the heading is the same mismatch this PR already fixed one file over, when README's "Option C: development installation" became "Option C: run from a checkout" for exactly this reason. Renaming it, or adding `--extra dev`, finishes the job. The file is outside this diff so there is no line to comment on, and a follow-up is fine.

**Verdict: APPROVE.**

junho226 (approved):
## Re-review at `53b21b6`

My previous review was dismissed by this push. The finding I raised is closed, the delta is docs-only, and every gate re-runs clean from a fresh checkout. **Approving.**

### The prior finding is closed

`8d2ecfc..53b21b6` touches two files, +6/-4, no `pyproject.toml`, no lockfile, no workflow:

| Site | Section | Now reads |
|---|---|---|
| `CONTRIBUTING.md:31-32` | Set up development environment | note + `uv sync --extra dev` |
| `docs/installation.md:393-394` | For contributing | note + `uv sync --extra dev` |
| `docs/installation.md:465` | Update installation | `uv sync  # add --extra dev in a development checkout` |
| `docs/installation.md:522` | Virtual environment issues | `uv sync  # add --extra dev in a development checkout` |

L465 and L522 are the two recipes I named, and the setup-block note covers the general direction. I reproduced the behavior at this HEAD rather than trusting the note:

```
$ uv sync                       # immediately after uv sync --extra dev --locked
Resolved 48 packages in 5ms
Uninstalled 12 packages in 170ms
 - coverage==7.13.4      - mypy==1.18.2          - pytest==9.0.3
 - pytest-asyncio==1.3.0 - pytest-cov==7.0.0     - ruff==0.15.13
 ... (iniconfig, mypy-extensions, packaging, pathspec, pluggy, pygments)

$ uv run --no-sync pytest --version
error: Failed to spawn: `pytest`
```

The note's scope is right: `--extra dev` restores exactly what a bare sync removes. `ruff` appears in that uninstall list only because I had installed it into the venv the way `test.yml:46` does; the documented contributor path is `uv tool install ruff==0.15.13` (`CONTRIBUTING.md:37`), which lives outside the venv and no `uv sync` touches. Keeping the flag a trailing note in the two recipes is also the right call, since `systemctl restart` / `docker-compose restart` sit in the same blocks.

### Nothing was dropped but the dead tools

Parsed both `pyproject.toml` revisions with `tomllib` and compared the union of the two old lists against the surviving one, rather than reading the TOML diff:

| Package | `[optional-dependencies] dev` before | `[dependency-groups] dev` before | after |
|---|---|---|---|
| pytest | `>=7.0.0` | `>=8.4.2` | `>=8.4.2` |
| pytest-asyncio | `>=0.21.0` | `>=1.2.0` | `>=1.2.0` |
| pytest-cov | absent | `>=6.0.0` | `>=6.0.0` |
| mypy | `>=1.0.0` | absent | `>=1.15.0` |
| black | `>=23.0.0` | absent | dropped |
| isort | `>=5.12.0` | absent | dropped |
| flake8 | `>=6.0.0` | absent | dropped |

`dropped: ['black', 'flake8', 'isort']`, `added: []`, every survivor keeps the higher of the two floors, and `[project.dependencies]` is byte-identical. The three removals are dead: no `[tool.black]`, no `[tool.isort]`, no `.flake8` / `setup.cfg` / `tox.ini` / `Makefile` / `noxfile.py`, and a multiline-safe `rg -U -i 'black|isort|flake8'` over the tree (minus `uv.lock`) returns only ruff's own selectors and `[tool.ruff.lint.isort]`. `rg -U 'dependency-groups|--group|--all-groups|--no-default-groups'` returns nothing at all, so no workflow or doc still names the removed group.

The surviving layout is a PEP 621 extra, which is what both consumers actually read — `uv sync --extra dev` and `pip install -e .[dev]`. A PEP 735 group would have been reachable by neither.

### Lockfile is regenerated and consistent

`uv lock --check` exits 0. Diffing the two `uv.lock` revisions structurally: 52 packages to 44, removed exactly `black, flake8, isort` plus their transitive closure `mccabe, platformdirs, pycodestyle, pyflakes, pytokens`, **added nothing, and changed no version of anything else**. `[package.metadata.requires-dev]` is gone and the locked dev extra is now `mypy, pytest, pytest-asyncio, pytest-cov` — matching `pyproject.toml` exactly, which is what `--locked` in CI enforces. Every locked version clears its new floor (pytest 9.0.3, pytest-asyncio 1.3.0, pytest-cov 7.0.0, mypy 1.18.2).

### Every consumer still resolves

- `.github/workflows/test.yml` is the only thing that installs the dev list. The two ad-hoc installs it retires (`uv pip install pytest-cov`, `uv pip install mypy`) are both now declared in the extra, so nothing quietly stops being installed. Ruff is pinned in at L46 after the sync and every later step passes `--no-sync`, so nothing prunes it.
- `.pre-commit-config.yaml` builds isolated envs and reads no project metadata; the `mirrors-mypy` rev `v1.15.0` matches the new `mypy>=1.15.0` floor, and the ruff rev `v0.15.13` matches the literal in `test.yml:25`, `CONTRIBUTING.md:37`, and `docs/installation.md:400`.
- `publish.yml` uses `build`/`twine`, `docker.yml` installs no Python deps, and the `Dockerfile` installs a prebuilt wheel. None reads the extra or the group.
- The comment justifying the mypy floor checks out: `test.yml:54` is `continue-on-error: true`, so that step is advisory and need not match the pre-commit rev exactly.

Dropping the `uv venv` step is fine — `uv sync` creates `.venv` under the resolved interpreter itself, which is what happened on my clean clone.

### Verification

Fresh clone at `53b21b6`, no pre-existing `.venv`, installed the way CI does. uv 0.11.21, CPython 3.12.9.

```
$ uv lock --check
Resolved 48 packages in 3ms                                     # exit 0

$ uv sync --extra dev --locked
+ pytest==9.0.3  + pytest-asyncio==1.3.0  + pytest-cov==7.0.0   # exit 0
$ uv run --no-sync mypy --version
mypy 1.18.2 (compiled: yes)

$ uv pip install ruff==0.15.13 && uv run --no-sync ruff check .
All checks passed!                                              # exit 0

$ uv run --no-sync ruff format --check .
82 files already formatted                                      # exit 0

$ uv run --no-sync mypy --ignore-missing-imports --no-strict-optional .
Success: no issues found in 82 source files                     # exit 0

$ PYTHONPATH=. uv run --no-sync pytest tests/ --tb=short --cov=. \
    --cov-report=term-missing --cov-fail-under=60
TOTAL                                               12845    639    95%
Required test coverage of 60% reached. Total coverage: 95.03%
1127 passed, 7 warnings in 17.72s                               # exit 0
```

The `.gitignore` addition covers all three patterns — `git check-ignore -v` resolves `.coverage` to `.gitignore:11`, `.coverage.abc123` to `:12`, `htmlcov/index.html` to `:13` — and `git status --porcelain` is empty after that full run. `CONTRIBUTING.md:139` is the `--cov-report=html` that motivates the third.

**Not run locally:** the 3.13 and 3.14 matrix legs (mine is 3.12.9 only) and the dropped `uv venv` step on a GitHub runner — only CI can confirm those, and all six checks are green at this HEAD, `test (3.12)`, `test (3.13)`, `test (3.14)` included. Also not run: `pre-commit run --all-files`, since no workflow runs pre-commit and it is not a merge gate here; the revs it pins are covered by the version-consistency check above.

### One optional follow-up, not a finding

`docs/getting-started.md:82` is headed "Method 3: Development setup" and installs with a bare `uv sync`, so it now hands you no test runner. Nothing in that file runs the tests, so no documented flow breaks, and your setup-block note covers the contributor who reaches for a bare sync. Worth a rename only because this PR already made exactly that fix one file over, turning README's "Option C: development installation" into "Option C: run from a checkout". The file is outside this diff, so a follow-up is fine.

### What lasts

`--locked` turns lockfile drift into a build failure and retires two ad-hoc installs in the same stroke. And the correction went past the finding: patching the two recipes closes the traps I found, while the setup-block note covers the ones nobody enumerated.

**Verdict: approve.**

